PR #12, opened 2018-08-07
Reorg Data.Macaw.Memory

karljs:
Moved the majority of the `Data.Macaw.Memory` module into `Data.Macaw.Memory.Common`, and changed the former to re-export useful stuff.


joehendrix:
I think this change is mostly an improvement, but keep changing the Memory files, and so want to hold off until it is more stable.

PR #28, opened 2019-03-04
Change an error call to a safe failure

travitch:
@dmwit - I have a large collection of binaries that trigger this error condition consistently (I can send them to you).  Is it safe to just have this be a `Left` instead of a call to error?  I assume so, but want to check.


dmwit:
It is probably safe, but I'd lightly recommend against it -- it is almost certainly a bug in the calling code to pass a negative number there. Please do send me a binary or two to investigate; attaching it either here or on mattermost should work fine.

dmwit:
b6778a80c15c91e8a12eb5d0c9a65c309f664a46 has an alternate fix.

travitch:
Thanks.  I like the alternative fix for now.  I think I'll come back to this idea and see if we can come up with a sufficiently satisfying non-fatal way to surface these issues.

PR #39, opened 2019-04-18
Memory examples

kquick:
Adds builds of the example programs to the cabal file to ensure that the examples remain up-to-date.
Also fixes the memory example for the additional argument to `mapRegionPointers`.

### DIFF
--- a/symbolic/examples/memory.hs
+++ b/symbolic/examples/memory.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
+{-# OPTIONS_GHC -Wno-unused-top-binds #-}
 
 import           GHC.TypeLits
 import           Control.Monad.ST ( stToIO, RealWorld )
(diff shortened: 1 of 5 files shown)


travitch:
I mostly only added the examples to get syntax highlighting while I wrote the examples in Data.Macaw.Symbolic.  I'm not opposed to this change (obviously), but if we do it, we might want to just combine them into one file to save some linking time.

Alternatively, we could just get rid of them and I can turn the versions in the haddocks into doctests.  This set will be smaller, since each only has one or two declarations.

joehendrix:
I think it'd be worth having examples to help people understand how to use macaw-symbolic, but could we add comments about what each example does?  The doctest route seems good to me to.

joehendrix:
Is this PR still relevant?

kquick:
Thanks for the reminder: I'll look at implementing doctests with explanations to resolve this PR.

travitch:
Just a note: doctest is currently broken with the latest cabal versions for really dumb reasons.  I'm not sure if we want to rely on it, as it is fragile (despite me being the one who was originally agitating for doctest)

kquick:
The doctest route is now implemented along with the associated changes in PR #107

PR #43, opened 2019-04-30
Drop ite support

joehendrix:


travitch:
I think this looks good.  Can we just get a comment in the merge commit on the motivation for the change from `ParsedIte` to `ParsedBranch`?  It looks like they are morally similar, but `ParsedBranch` has a structure more consistent with the rest of the block terminators.  It seems equally expressive - is that true?

joehendrix:
`ParsedBranch` is less expressive from a theoretical point of view, but in practice likely expressive enough .  `ParsedIte` can be nested, so you could use it to create n-ary branches (among other things) whereas `ParsedBranch` only allows binary.

The main advantage of `ParsedBranch` is that one no longer has to recursively evaluate `StatementList` and there is a linear-sequence of Macaw code associated with each basic block.

PR #55, opened 2019-07-11
Add mhnf rewriting.

kquick:
The parseFetchAndExecute in Discovery attempts to identify ParsedITE
terminal statements by examining the value of the ip_reg via
valueAsApp and pattern matching on a Mux statement.  This patch adds
specific handling in the Macaw CFG Rewriter to attempt to float Mux
statements upwards (aka "Mux Head Normal Form") so that they will be
the top-most ip_reg value and therefore be recognized as a ParsedITE
terminator.

For macaw-ppc testing of the 988KB gzip binary, this increased the
number of blocks found from 1339 to 37950 (and increased the test
runtime from 1.36s to 88.14s).

### DIFF
--- a/base/src/Data/Macaw/CFG/Rewriter.hs
+++ b/base/src/Data/Macaw/CFG/Rewriter.hs
@@ -612,7 +612,54 @@ rewriteApp app = do
        else
         rewriteApp (Eq x (BVValue u (toUnsigned u yc)))
 
-    _ -> evalRewrittenRhs (EvalApp app)
+    -- no normal rewrites available, now try mhnf for enabling Discovery
+    _ -> rewriteMhnf app
+
+
+-- | Performs rewrites for "Mux Head Normal Form", which attempts to
+-- raise Mux operations upwards to make them more likely to appear as
+-- the last statement in a Block.  This is needed because
+-- parseFetchAndExecute in Discovery will attempt to recognize branch
+-- statements by pattern matching the value of the IP register as a
+-- Mux.
+--
+-- For example, the following:
+--
+--     r32 := Mux r31 (0x4 :: [64]) (0x28 :: [64])
+--     r33 := Add r32 (0x100a3e50)
+--     { ip => r33, ... }
+--
+-- should be rewritten to:
+--
+--     r34 := Mux r31 (0x100a3e54) (0x100a3e78)
+--     { ip => r34, ... }
+--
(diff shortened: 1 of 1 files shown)


kquick:
The comment about "this is done last after other rewriting" is not strictly applicable since the existing implementation is fully recursive, but the importance of lifting Mux operations to facilitate Discovery makes it worth emphasizing, which is the intent of the comment.

Also, thanks to WRHarris for the "mhnf" nomenclature, which is also not strictly correct, but close enough and distinctive enough to uniquely identify this concept.

joehendrix:
I guess this is merged, but for posterity why did you decide to implement this in the rewriter rather than extend `parseFetchAndExecute`?

kquick:
This needed to be done in the rewriter because in order to convert

    r37 := mux r36 x y
    r38 := bvadd r37 z

it needs to have the nonce generator available to generate another register.  Even if `z` is a `valueAsMemAddress` constant, the `ParsedITE` form requires the `t` and `f` branch to be a Core Value, so if `x` and `y` are not a `BVValue` or a `RelocatableValue`, then an `AssignedValue App` is needed, and thus the nonce requirement.  

Additionally, adding what is essentially term-rewriting logic to `parseFetchAndExecute` seemed like duplication of effort that the rewriter is already handling.

If there's something I've overlooked or there's a better way to handle this @joehendrix , I'm happy to work out a different approach and replace it in the tree, but because of the What4 Semiring changes in the Expr syntax, that floats out the constant  values and resulted in the Mux-then-Add shown above and essentially broke all ParsedITE recognition for macaw-semmc-based operations.

joehendrix:
I don't fully see the need to rewrite the above expression when `x` and `y` are not intra-procedural jump targets (e.g. relocatable values or bv values), but you have more context than I.
I'd personally rather code kept close to where it is needed (e.g., in `parseFetchAndExecute` or maybe even in the macaw-semmc code), and this rewrite seems tied to specific needs of `parseFetchAndExecute` and particular choices made by `What4`.  Other use cases may prefer a different normal form.  For example, a presentation I saw the over the weekend rewrote if-then-elses in the opposite way (see the slides on "Shared Monomial Lifting" at http://smt2019.galois.com/slides/slides_9.pdf).
If there's some regression tests though that'll complain if this breaks, then I guess I'd propose just keeping this as the code apparently is working.

PR #56, opened 2019-07-11
Semiring upd

kquick:
More updates for semiring What4 Expr changes, including macaw-ppc.


travitch:
Thanks, I think it looks good to me.  Unfortunately I'm not familiar with the semiring stuff, but I think I get the idea from this.

Issue #58, opened 2019-07-11
Update macaw-semmc conditionalBranch for removal of branch labels


kquick:
It is likely that conditionalBranch is no longer needed as part of the ppc and arm disassembly/Block generation operations, although for PPC the trap conditions will need to be turned into architecture specific extensions.

travitch:
Closed by #66

PR #66, opened 2019-08-09
Wip/ppc no block labels

travitch:
This is a draft of an update to macaw-ppc to accommodate the recent changes in macaw (i.e., removing block labels)

### DIFF
--- a/macaw-ppc/src/Data/Macaw/PPC/Disassemble.hs
+++ b/macaw-ppc/src/Data/Macaw/PPC/Disassemble.hs
@@ -143,15 +143,6 @@ disassembleBlock lookupSemantics gs curIPAddr blockOff maxOffset = do
             addStmt (InstructionStart blockOff (T.pack lineStr))
             addStmt (Comment (T.pack  lineStr))
             asAtomicStateUpdate (MM.segoffAddr curIPAddr) transformer)
-
-            -- Check to see if the IP has become conditionally-defined (by e.g.,
-            -- a mux).  If it has, we need to split execution using a primitive
-            -- provided by the Generator monad.
-            -- nextIPExpr <- getCurrentIP
-            -- case matchConditionalBranch nextIPExpr of
-            --   Just (cond, t_ip, f_ip) ->
-            --     conditionalBranch cond (setRegVal PPC_IP t_ip) (setRegVal PPC_IP f_ip)
-            --   Nothing -> return ())
           case egs1 of
             Left genErr -> failAt gs off curIPAddr (GenerationError i genErr)
             Right gs1 -> do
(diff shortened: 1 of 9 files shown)


macaw-ppc/src/Data/Macaw/PPC/Disassemble.hs
```diff
@@ -225,12 +199,12 @@ disassembleFn :: (ppc ~ SP.AnyPPC var, PPCArchConstraints var)
...
               -> Int
               -- ^ Maximum size of the block (a safeguard)
-              -> ST s ([Block ppc ids], Int, Maybe String)
+              -> ST s (Block ppc ids, Int)
 disassembleFn _ lookupSemantics nonceGen startAddr regState maxSize = do
   mr <- ET.runExceptT (unDisM (tryDisassembleBlock lookupSemantics nonceGen startAddr regState maxSize))
   case mr of
-    Left (blocks, off, exn) -> return (blocks, off, Just (show exn))
-    Right (blocks, bytes) -> return (blocks, bytes, Nothing)
+    Left (blocks, off, _exn) -> return (blocks, off)
```
travitch: I'd like to find a way to propagate this exception information up (assuming there is ever useful information there).  I'm actually not sure right now what could error out.

kquick: We could do it inline by adding a new Block type that essentially just contained the address and the error.  Then any analysis that attempted to execute that block could deliver a more focused result.

We can also just log it (when the logging gets here).

macaw-semmc/src/Data/Macaw/SemMC/Generator.hs
```diff
@@ -212,6 +176,10 @@ data GeneratorError = InvalidEncoding
                     | GeneratorMessage String
   deriving (Show)
 
+data PartialBlock arch ids where
+  UnfinishedPartialBlock :: !(PreBlock arch ids) -> PartialBlock arch ids
+  FinishedPartialBlock :: !(Block arch ids) -> PartialBlock arch ids
+
```
kquick: I like this: being able to distinguish between finished and unfinished.

macaw-ppc/src/Data/Macaw/PPC/Disassemble.hs
kquick: Added https://github.com/GaloisInc/macaw/issues/67

kquick:
Also note that this PR addresses https://github.com/GaloisInc/macaw/issues/58

macaw-ppc/src/Data/Macaw/PPC/Disassemble.hs
joehendrix: Are you looking for Macaw to immediately print out an error message?  If not, you can terminate the block with `TranslationError` and look for that.  If you want more structured information, then you could create an architecture-specific terminal statement.
I removed the error message because the block itself contained the error in the form of a translation error.

travitch: Actually, looking into it the implementation of `failAt` already turns the thrown exception into a `TranslationError` and that reported exception is extraneous.

PR #79, opened 2019-10-22
Division changes

joehendrix:
This combines unsigned/signed quotient and remainder into a single operation for each signed/unsigned case.


travitch:
Is the summary here that div/rem are now combined syntactic forms instead of standalone operations?

joehendrix:
That's a good summary.  The backstory is that John Regehr found an error in Reopt translation from the existing operations to LLVM, and finding some other issues, I think it's best if I have a single operation.

I'm also ensuring macaw-x86-symbolic has proper overflow/divide-by-zero checks.

PR #95, opened 2019-11-27
Change the BlockClassifierContext in Discovery into a datatype (from …

travitch:
…a tuple)

@joehendrix Do you have any objections to this change?

I'm about to add a new field, and the 7-tuple was a bit confusing. I wanted to make this a separate commit so that we could discuss it independent from the other change.


joehendrix:
Looks fine to me.  With the datatype, I think all types could be made strict.

travitch:
What benefit do you see for strictness? In this case, only one instance of the type is created (accessed many times). I could imagine that strictness would prevent anyone from accidentally forgetting to initialize a field, which is a reasonable benefit.  If that is it, I'm happy to make the change.

joehendrix:
I just tend to default strict records unless I am deliberately trying to be lazy.  Generally, it prevents the uninitialized fields as you say, also typically leads to clearer tracing, and prevents unintended variable capture in unevaluated closures.   It's not a big deal though in this case.

PR #98, opened 2019-12-16
Introduce separate stack-analysis algorithm.

joehendrix:


joehendrix:
Migrating to staging approach with jhx/develop

PR #100, opened 2019-12-19
Add a new entry point to macaw-symbolic

travitch:
This version constructs a Crucible CFG for a collection of blocks while
preserving control flow between them.  It allows the caller to specify blocks
that are considered "terminal": those blocks return the current register state.
Control flow to blocks no included in the "slice" are directed to synthetic
blocks that assume False in order to stop the symbolic simulator from exploring
those branches.


travitch:
This is now merged into #118 (which will be merged soon)

PR #107, opened 2020-02-06
Tr/rewrite symbolic memory model

travitch:
This is a revision of the macaw-symbolic memory model

It fixes a soundness bug in the core, and provides a much more efficient representation in the recommended memory model.


travitch:
The commit message for the squashed commit will be roughly:

This commit re-implements the memory model used by macaw symbolic

# Summary

There are two major changes:
- The interface to memory models in Data.Macaw.Symbolic has changed
- The suggested implementation in Data.Macaw.Symbolic.Memory has changed

The change improves performance and fixes a soundness bug.

## Migration Guide

* `macawExtensions` (Data.Macaw.Symbolic) takes a new argument: a `MkGlobalPointerValidityPred`.  Use `mkGlobalPointerValidityPred` to provide one.
* `mapRegionPointers` no longer takes a default pointer argument (delete it at call sites)

# Details

## Rationale

This interface and implementation is concerned with handling pointers to static
memory in a binary.  These are distinguished from pointers to
dynamically-allocated or stack memory because many machine code instructions
compute bitvectors and treat them as pointers.  In the LLVM memory model used by
macaw-symbolic, each memory allocation has a block identifier (a natural
number).  The stack and each heap allocation get unique block identifiers.
However, pointers to static memory have no block identifier and must be mapped
to a block in order to fit into the LLVM memory model.

The previous implementation assigned each mapped memory segment in a binary to
its own LLVM memory allocation.  This had the nice property of implicitly
proving that no memory access was touching unmapped memory.  Unfortunately, it
was especially inefficient in the presence of symbolic reads and writes, as it
generated mux trees over all possible allocations and significantly slowed
symbolic execution.

## Implementation

The new memory model implementation (in Data.Macaw.Symbolic.Memory) instead uses
a single allocation for all static allocations.  This pushes more of the logic
for resolving reads and writes into the SMT solver and the theory of arrays.  In
cases where sufficient constraints exist in path conditions, this means that we
can support symbolic reads and writes.  Additionally, since we have only a
single SMT array backing all allocations, mapping bitvectors to LLVM pointers in
the memory model is trivial: we just change their block identifier from zero
(denoting a bitvector) to the block identifier of the unique allocation backing
static data.

This change has to do some extra work to ensure safety (especially that unmapped
memory is never written to or read from).  This is handled with the
MkGlobalPointerValidityPred interface in Data.Macaw.Symbolic.  This function,
which is passed to the macaw-symbolic initialization, constructs well-formedness
predicates for all pointers used to access memory.  Symbolic execution tasks
that do not need to enforce this property can simply provide a function that
never returns any predicates to check.  Implementations that want a reasonable
default can use the mkGlobalPointerValidityPred from Data.Macaw.Symbolic.Memory.
The default implementation ensures that no reads or writes touch unmapped memory
and that writes to static data never write to read-only segments.

This change also converts the examples in macaw-symbolic haddocks to use doctest
to ensure that they do not get out of date.  These are checked as part of CI.

symbolic/src/Data/Macaw/Symbolic/Memory.hs
```diff
@@ -314,30 +371,97 @@ pleatM :: (Monad m, F.Foldable t)
...
+  -- return False for the predicate on that range).
+  --
+  -- Otherwise, the pointer is allowed to be between the lo/hi range
+  let inMappedRange off (range, mut)
+        | MS.pointerUseTag puse == MS.PointerWrite && mut == CL.Immutable = return (WI.falsePred sym)
+        | otherwise =
+          case range of
+            IM.IntervalCO lo hi -> do
+              lobv <- WI.bvLit sym (memRepr mpt) (fromIntegral lo)
+              hibv <- WI.bvLit sym (memRepr mpt) (fromIntegral hi)
```
kquick: The `lobv` and `hibv` are identical for each case and could be computed prior to the `case`.

symbolic/src/Data/Macaw/Symbolic/Memory.hs
```diff
@@ -145,37 +154,37 @@ data MemoryModelContents = SymbolicMutable
...
 -- | An index of all of the (statically) mapped memory in a program, suitable
 -- for pointer translation
 data MemPtrTable sym w =
-  MemPtrTable { memPtrTable :: IM.IntervalMap (MC.MemWord w) (Allocation sym w)
-              , allocationIndex :: Map.Map (CS.RegValue sym CT.NatType) (Allocation sym w)
+  MemPtrTable { memPtrTable :: IM.IntervalMap (MC.MemWord w) CL.Mutability
+              -- ^ The ranges of (static) allocations that are mapped
+              , memPtr :: CL.LLVMPtr sym w
+              -- ^ The pointer to the allocation backing all of memory
+              , memRepr :: PN.NatRepr w
```
kquick: Document this as the mem word width for this allocation.

symbolic/src/Data/Macaw/Symbolic/Bitcast.hs
```diff
@@ -0,0 +1,94 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeOperators #-}
+module Data.Macaw.Symbolic.Bitcast (
```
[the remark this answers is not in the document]
travitch: Correct, I was diagnosing an error that prevented doctest from being able to run. It turned out that it was related to `-XStarIsType`, which conveniently only occured with the type level multiplications here.

symbolic/src/Data/Macaw/Symbolic/Memory.hs
travitch: Unfortunately, there are no extractors for the `Interval` type to get at the bounds.  I could write one here to factor those out. Do you think it is worth it, given that they need to pattern match anyway?

kquick: Ahh, that makes sense.  No, I don't think it's worth it in that case.

PR #108, opened 2020-02-07
Update register usage for reopt

joehendrix:


base/src/Data/Macaw/Memory.hs
```diff
@@ -1060,7 +1060,7 @@ instance Hashable (MemAddr w) where
   hashWithSalt s a = s `hashWithSalt` addrBase a  `hashWithSalt` addrOffset a
 
 instance Show (MemAddr w) where
-  showsPrec _ (MemAddr 0 a) = showString "0x" . shows a
+  showsPrec _ (MemAddr 0 a) = shows a
```
travitch: I was wondering where that extra `0x` was coming from

joehendrix: Addresses need lots of hugs and kisses.

PR #109, opened 2020-02-07
Macaw updates for Reopt register usage analysis.

joehendrix:


base/src/Data/Macaw/Analysis/RegisterUse.hs
```diff
@@ -352,191 +1315,316 @@ requiredAssignDeps :: OrdF (ArchReg arch)
...
+      cns   <- gets blockUsageStartConstraints
+      cache <- gets assignDeps
+      let deps = mconcat
+                    [ assignSet aid
+                    , valueDeps cns cache cond
+                    , valueDeps cns cache addr
+                    , valueDeps cns cache val
+                    ]
+      requiredAssignDeps aid deps
+      addDeps $ assignSet aid
```
travitch: Why does this (`CondReadMem`) case get this call to `addDeps` while the analogous case in the `ReadMem` handler above does not?

base/src/Data/Macaw/Analysis/RegisterUse.hs
```diff
@@ -352,191 +1315,316 @@ requiredAssignDeps :: OrdF (ArchReg arch)
...
+     accessInfo <- popAccessInfo
+     case accessInfo of
+       NotFrameAccess -> do
+         demandValue c
+         demandValue addr
+         demandValue val
+       FrameReadInitAccess{} -> error "Expected conditional write access"
+       FrameReadWriteAccess{} -> error "Expected conditional write access"
+       FrameReadOverlapAccess{} -> error "Expected conditional write access"
+       FrameWriteAccess{} -> error "Expectedf1 conditional write access"
```
travitch: Typo in error mesage

base/src/Data/Macaw/Analysis/RegisterUse.hs
```diff
@@ -352,191 +1315,316 @@ requiredAssignDeps :: OrdF (ArchReg arch)
...
+      ParsedArchTermStmt tstmt regs _mnext -> do
+        summaryFn <- asks reguseTermFn
+        s <- get
+        case summaryFn tstmt regs s of
+          Left emsg -> throwError emsg
+          Right rDeps -> blockRegDependenciesLens .= rDeps
+      ParsedTranslateError _ ->
+        error "Cannot identify register use in code where translation error occurs"
+      ClassifyFailure _ _ ->
+        error $ "Classification failed: " ++ show addr
```
travitch: Would it be appropriate to turn these into `throwError` calls instead?

base/src/Data/Macaw/Analysis/RegisterUse.hs
```diff
@@ -567,53 +1658,151 @@ backPropagateOne s rest newLocs ((srcAddr,srcDepMap):predRest) = do
...
+                     , ShowF (ArchReg arch)
+                     , MemWidth (ArchAddrWidth arch)
+                     )
+                  => FunPredMap (ArchAddrWidth arch)
+                  -> (ArchSegmentOff arch
+                       -> ArchSegmentOff arch
+                       -> PostValueMap arch ids)
+                     -- ^ Maps block address to map of location representative to value.
+                  -> ArchSegmentOff arch
+                     -- ^ Address of thsi block.
```
travitch: thsi -> this

base/src/Data/Macaw/Analysis/RegisterUse.hs
```diff
@@ -567,53 +1658,151 @@ backPropagateOne s rest newLocs ((srcAddr,srcDepMap):predRest) = do
...
+                  .  (HasRepr (ArchReg arch) TypeRepr
+                     , OrdF (ArchReg arch)
+                     , ShowF (ArchReg arch)
+                     , MemWidth (ArchAddrWidth arch)
+                     )
+                  => FunPredMap (ArchAddrWidth arch)
+                  -> (ArchSegmentOff arch
+                       -> ArchSegmentOff arch
+                       -> PostValueMap arch ids)
+                     -- ^ Maps block address to map of location representative to value.
```
travitch: Is the comment implying that I should think of this as a function where I apply a block address to get a (map of location representatives to values)?

base/src/Data/Macaw/Analysis/RegisterUse.hs
```diff
@@ -567,53 +1658,151 @@ backPropagateOne s rest newLocs ((srcAddr,srcDepMap):predRest) = do
...
+         , biUsedWriteSet  = dsWriteStmtIndexSet deps
+         , biMemAccessList = blockMemAccesses summary
+         , biLocMap = mkDepLocMap cns
+         , biPredPostValues = Map.fromList predphilist
+         , biPhiLocs = Set.toList (dsLocSet deps)
+         , biStartConstraints = cns
+         }
+
 -- | This analyzes a function to determine which registers must b available to
 -- the highest index above sp0 that is read or written.
```
travitch: b -> be

base/src/Data/Macaw/Analysis/RegisterUse.hs
```diff
@@ -5,22 +5,51 @@ task needed before deleting unused code.
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE UndecidableInstances #-}
 module Data.Macaw.Analysis.RegisterUse
```
travitch: Module level comments/questions:
- A high-level description of the intent would be useful, or maybe just additional commentary on the `registerUse` function. My interpretation of this module is that it discovers what registers and memory locations must be populated for the input function to have defined behavior (ish).
- From the representation I think I see, this only talks about "top level" register and memory structure - it doesn't say anything about e.g., a pointer to a struct having particular fields initialized
- It seems like this code makes assumptions about calls obeying the ABI - is it the responsibility of the caller to validate that?
- It would be really useful to have a test or two in macaw-x86 to make sure that nobody else breaks anything

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -636,345 +147,141 @@ joinInitialBounds :: forall arch
...
-          discardStackInfo bnds
+      case intraStackValueExpr cns addr of
+        StackOffsetExpr addrOff ->
+          writeStackOff cns addrOff repr val
+        -- If we write to something other than stack, then clear all
+        -- stack references.
+        --
+        -- This is probably not a good idea in general, but seems fine
+        -- for stack detection.
+        _ ->
```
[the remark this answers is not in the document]
joehendrix: Basically, the idea is that typically if they do, the analysis results may not be correct, but that doesn't really matter for jump bounds detection.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -636,345 +147,141 @@ joinInitialBounds :: forall arch
...
--- | Discard information about the stack in the bounds due to an
--- operation that may affect the stack.
-discardStackInfo :: IntraJumpBounds arch ids
-                 -> IntraJumpBounds arch ids
-discardStackInfo bnds = bnds { stackExprMap = emptyStackMap }
+instance ShowF (ArchReg arch) => Pretty (IntraJumpBounds arch ids) where
+  pretty = pretty . initJumpBounds
+
+
+modifyIntraStackConstraints :: (BlockIntraStackConstraints arch ids -> BlockIntraStackConstraints arch ids)
```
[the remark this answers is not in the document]
joehendrix: It could be.

PR #113, opened 2020-02-28
What4 is now a separate repository from crucible.

kquick:


kquick:
No longer needed, these changes were effected and merged in #112

PR #119, opened 2020-03-20
Fix register-use; add null terminated string reader to Memory.

joehendrix:


joehendrix:
I did a brief check and it looks like the briareus fails are unrelated to this, so I'm going to go ahead and merge.

PR #126, opened 2020-04-13
Small fixes to macaw-symbolic tests for recent changes.

kquick:


travitch:
Whoops, thanks.

One of these days, Real Soon Now, I'm going to go fix all of the star is type stuff...

PR #130, opened 2020-04-18
Clean up the PowerPC architecture specifications

travitch:
This commit reduces duplication by defining some instances in terms of the
parameterized AnyPPC architecture.  It also removes some code that was marked as
deprecated, now that we can use the infrastructure built up around AnyPPC.


travitch:
This is a WIP - I still need to update macaw-ppc-symbolic

PR #132, opened 2020-05-04
Fixed buggy rewrite rules involving testBit/shifts

benjaminselfridge:
I think all three of the shifts were erroneous in different ways -- 'BVTestBit' should be 0 when the index is larger than the width of the word. 

Edit: I was wrong; the only bug was the logical shift left case.

### DIFF
--- a/base/src/Data/Macaw/CFG/Rewriter.hs
+++ b/base/src/Data/Macaw/CFG/Rewriter.hs
@@ -507,13 +507,17 @@ rewriteApp app = do
     -- (x << j) testBit i ~> x testBit (i-j)
     -- plus a couple special cases for when the tested bit falls outside the shifted value
     BVTestBit (valueAsApp -> Just (BVShr w x (BVValue _ j))) (BVValue _ i)
-      | j + i <= maxUnsigned w -> do
+      | j + i < intValue w ->
       rewriteApp (BVTestBit x (BVValue w (j + i)))
+      | otherwise -> pure (boolLitValue False)
     BVTestBit (valueAsApp -> Just (BVSar w x (BVValue _ j))) (BVValue _ i)
-      | i < intValue w -> do
-      rewriteApp (BVTestBit x (BVValue w (min (j + i) (intValue w-1))))
+      | j + i < intValue w ->
+      rewriteApp (BVTestBit x (BVValue w (j + i)))
+      | i < intValue w -> pure (boolLitValue True)
+      | otherwise -> pure (boolLitValue False)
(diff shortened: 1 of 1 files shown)


base/src/Data/Macaw/CFG/Rewriter.hs
```diff
@@ -507,13 +507,17 @@ rewriteApp app = do
...
+      | j + i < intValue w ->
       rewriteApp (BVTestBit x (BVValue w (j + i)))
+      | otherwise -> pure (boolLitValue False)
     BVTestBit (valueAsApp -> Just (BVSar w x (BVValue _ j))) (BVValue _ i)
-      | i < intValue w -> do
-      rewriteApp (BVTestBit x (BVValue w (min (j + i) (intValue w-1))))
+      | j + i < intValue w ->
+      rewriteApp (BVTestBit x (BVValue w (j + i)))
+      | i < intValue w -> pure (boolLitValue True)
+      | otherwise -> pure (boolLitValue False)
```
dmwit, line 517: I think these two cases aren't quite right: they should both pick out the top bit of x.

dmwit: Or, hm. Maybe the first one should pick out the top bit and the second one is fine? I don't know what the intended behavior of testBit for out-of-range indices is.

base/src/Data/Macaw/CFG/Rewriter.hs
```diff
@@ -507,13 +507,17 @@ rewriteApp app = do
...
     BVTestBit (valueAsApp -> Just (BVShl w x (BVValue _ j))) (BVValue _ i)
-      | j <= i -> rewriteApp (BVTestBit x (BVValue w (i - j)))
+      | 0 <= i - j && i <= intValue w ->
+      rewriteApp (BVTestBit x (BVValue w (i - j)))
```
dmwit, line 520: I think this extra condition will never be true; it's already covered by line 479.

base/src/Data/Macaw/CFG/Rewriter.hs
benjaminselfridge: You are right -- the second one should pick out the top bit. Let me see if I can fix it.

base/src/Data/Macaw/CFG/Rewriter.hs
benjaminselfridge: You're right

benjaminselfridge:
@dmwit Thanks for the comments -- Can you take another look when you get a chance?

joehendrix:
I'd recommend squashing this to a single commit, but otherwise it looks fine to me.

dmwit:
LGTM

PR #136, opened 2020-05-15
Handle bitwise operations on stack offset abstract values

chameco:


travitch:
I believe this looks good.  Is the summary
```
- Generalize handling of bitwise operations to also apply them to stack offsets
- Use the extended bitwise handling on AND
```

It looks good in principle - do the macaw-x86, macaw-ppc, and macaw-aarch32 tests still pass?  I worry that we don't have as many tests for that code path as we want.  Could we check your motivating example for x86 in?

chameco:
@travitch That sums it up. I've added my example as a test, and confirmed that the `macaw-x86` and `macaw-ppc` tests pass. GHC runs out of memory on my machine when building `Data.Macaw.ARM.Semantics.ARMSemantics`, so I'm unsure if the tests for `macaw-aarch32` are passing.

travitch:
That looks good to me then.  Can we squash merge it and add that description as the message?

PR #137, opened 2020-05-15
Tr/full arm intrinsics

travitch:
This PR improves support for AArch32.  Most significantly it adds arch-specific functions to support all of the uninterpreted floating point operations.

- It avoids serializing and immediately deserializing the semantics for ARM while running TH (this is accompanied by a bunch of changes in asl-translator and semmc-aarch32).
- It changes the code generation strategy for all macaw-semmc architecture: all bindings are generated lazily where possible (using let bindings) and forced to preserve the ordering of side effects.  This lets us more easily (and uniformly) handle conditional execution of side effects while preserving sharing.
- The arguments to `genExecInstruction` are cleaned up with a configuration
- Currently the ARM vector instructions are represented with a single uninterpreted statement because generating all of them creates way too much code to actually compile


macaw-aarch32/src/Data/Macaw/ARM/Semantics/TH.hs
```diff
@@ -326,34 +588,50 @@ armAppEvaluator :: M.Endianness
...
     case elt of
       WB.BaseIte bt _ test t f | isPlaceholderType bt -> return $ do
+        -- NOTE: This case is very special.  The placeholder types denote
+        -- conditionals that are guarding the state update functions with
+        -- mutation.
+        --
+        -- We need to ensure that state updates are only done lazily.  This
+        -- works because the arguments to the branches are expressions in the
+        -- Generator monad.  We can do this translation while preserving sharing
+        -- by turning every recrusively-traversed term into a let binding at the
```
kquick, line 598: recursively

travitch: Thanks!

PR #138, opened 2020-05-18
Update submodules for parameterized-utils release 2.1.0 compatibility.

kquick:


kquick:
I am not using the stack files myself and have no objection to removing them.

My general position is one of agnosticism towards which build tool people prefer, so if there is someone that is interested in using stack I wouldn't want to deter them, but in this case the stack files are definitely broken, so I agree that it's better to remove them and anyone interested in maintaining them can resurrect them based on the contents at this point.

PR #156, opened 2020-08-05
aarch32: support non-concrete conditional writes

danmatichuk:


macaw-aarch32/src/Data/Macaw/ARM/Semantics/TH.hs
```diff
@@ -414,79 +419,275 @@ unitValue =  M.Initial ARMDummyReg
...
+                -> WriteSeq f addr val m
+                -> WriteSeq f addr val (n + m)
+appendWriteSeqs s1 s2 = case (s1, s2) of
+  (EmptyWriteSeq, _) -> s2
+  (_, EmptyWriteSeq) -> s1
+  (WriteSeq v1, WriteSeq v2)
+     | LeqProof <- leqAddPos (V.length v1) (V.length v2)
+     -> WriteSeq $ V.append v1 v2
+
+data ARMWriteActions ids addr val where
```
travitch, line 453: What are `addr` and `val`? I see them instantiated as 4/8/16 below.

macaw-aarch32/src/Data/Macaw/ARM/Semantics/TH.hs
```diff
@@ -414,79 +422,285 @@ unitValue =  M.Initial ARMDummyReg
 natReprFromIntTH :: Int -> Q Exp
 natReprFromIntTH i = [| knownNat :: M.NatRepr $(litT (numTyLit (fromIntegral i))) |]
 
+-- | A representation of an
+-- un-executed effectful generator action, where 'addr' is a generalized
+-- address (either a memory address, or a register number) and 'val' is the value
+-- to be written.
```
travitch, line 428: I think `addr` and `val` turned into `w` - is that just pointer width now?

macaw-aarch32/src/Data/Macaw/ARM/Semantics/TH.hs
```diff
@@ -414,79 +422,285 @@ unitValue =  M.Initial ARMDummyReg
...
+data WriteAction f w where
+  -- | A single write action 
+  WriteAction :: forall f w
+               . (1 <= w)
+              => M.NatRepr w
+              -> f M.BoolType
+              -- ^ a guard which indicates whether or not this action should be committed
+              -> f (M.BVType (8 TL.* w))
```
travitch, line 436: This is the value being written?  I assume that the location is now implicit because this will be run at a higher level where the destination of the write is apparent.

travitch: oh, `w` is the number of bytes

macaw-aarch32/src/Data/Macaw/ARM/Semantics/TH.hs
```diff
@@ -594,18 +799,18 @@ armAppEvaluator :: M.Endianness
...
-        -- a Mux.
+        -- rather than an actual macaw term. The mux condition is therefore mapped
+        -- across all of the stateful actions
         testE <- addEltTH endianness interps test
-        tE <- addEltTH endianness interps t
-        fE <- addEltTH endianness interps f
-        case all isEager [testE, tE, fE] of
-          True -> liftQ [| return $ concreteIte $(refEager testE) $(refEager tE) $(refEager fE) |]
-          False -> liftQ [| concreteIte <$> $(refBinding testE) <*> $(refBinding tE) <*> $(refBinding fE) |]
+        inConditionalContext $ do
```
travitch, line 808: Interesting that this came back

macaw-aarch32/src/Data/Macaw/ARM/Semantics/TH.hs
danmatichuk: Yeah, `w` is the number of bytes to be written, and `addr` is the width of the "address" (i.e. 32 bits for memory accesses, 4 bits for GPRs and 8 bits for SIMDs). It doesn't appear here now because it instead appears as the key to the map, mapping addresses to write actions.

macaw-aarch32/src/Data/Macaw/ARM/Semantics/TH.hs
danmatichuk: Yeah, we're now unconditionally executing both branches in the Generator monad and only making the writes themselves conditional. This ends up generating unnecessary memory reads for branches that don't get executed, so we need to wrap those reads up as lazily-evaluated values.

macaw-aarch32/src/Data/Macaw/ARM/Semantics/TH.hs
danmatichuk: Just `val`, see here: https://github.com/GaloisInc/macaw/pull/156#discussion_r466683933

PR #157, opened 2020-08-09
Update for addition of badBehaviorMap implicit for LLVM annotations.

kquick:
See https://github.com/GaloisInc/crucible/pull/453


robdockins:
It seems like this subsumes #153, is that correct?

kquick:
> It seems like this subsumes #153, is that correct?

Oops, I had not seen 153 before starting this effort. Sorry for the duplication, but this one is probably a superset of the changes and is ready to be merged.

PR #158, opened 2020-09-02
macaw-symbolic: Support alternative memory models in ArchInfo

danmatichuk:


symbolic/src/Data/Macaw/Symbolic.hs
```diff
@@ -185,30 +191,34 @@ import qualified Data.Macaw.Symbolic.MemTraceOps as MT
...
+      :: forall a m sym proxy
+       . (IsSymInterface sym, MemModelConstraint mem sym, MonadIO m)
+      => proxy mem
+      -> sym
+      -> (SB.MacawArchEvalFn sym (MemModelType mem arch) arch -> m a)
+      -> m a
+
 -- | The information to support use of macaw-symbolic for a given architecture
-data ArchVals arch = ArchVals
+data ArchVals arch = HasMemoryModel arch LLVMMemory => ArchVals
```
travitch, line 217: Is the consequence here that the only way we can symbolically simulate machine code is if we use the LLVM memory model?  How painful is it to restructure this to make the memory model a parameter to `ArchVals`?

joehendrix: Separately, does this mean to introduce the `HasMemoryModel` constraint you have to pattern match on the `ArchValsValue` constructor?  That seems like I would need a comment every time I did it to explain what was happening.  Why not just have a `withMemoryModel` constructor?

symbolic/src/Data/Macaw/Symbolic.hs
```diff
@@ -182,33 +188,45 @@ import qualified Data.Macaw.Symbolic.MemTraceOps as MT
...
+  :: ArchInfo arch
+  => proxy arch
+  -> Maybe (ArchVals arch)
+archVals p = genArchVals (Proxy @LLVMMemory) p
+
+data LLVMMemory
+
+class IsMemoryModel mem where
+  type MemModelType mem arch :: C.CrucibleType
+  type MemModelConstraint mem sym :: Constraint
```
travitch, line 210: It looked like this might be the only use of the `constraints` package - can we just import `Constraint` from `GHC.Exts` instead?  That would also get rid of the import of Data.Constraints and the extra dependency in the cabal file.

symbolic/src/Data/Macaw/Symbolic.hs
```diff
@@ -126,6 +133,8 @@ import           Control.Monad.IO.Class
 import qualified Data.BitVector.Sized as BV
 import qualified Data.Foldable as F
 import qualified Data.Map.Strict as Map
+import           Data.Constraint
+import           Data.Proxy
```
travitch, line 137: Is this import used anymore?

travitch (commented):
Thanks, I like this simplification

PR #159, opened 2020-09-11
Update to use new `HasLLVMAnn` API

robdockins:
It now requires an action for recording (or discarding) annotations rather than a map.


kquick (approved):
Looks straightforward.  The implication is that macaw doesn't preserve any LLVM annotation info and we'd change this in the future if we wanted some.

robdockins:
> Looks straightforward. The implication is that macaw doesn't preserve any LLVM annotation info and we'd change this in the future if we wanted some.

Exactly.

PR #161, opened 2020-09-29
Switch to GitHub actions

joehendrix:


kquick:
My only comment is that I think a could of debug echo statements got left behind, but they shouldn't hurt anything.

joehendrix:
The explicit echo statements are there to update the path for later commands.

PR #165, opened 2020-10-05
symbolic: Add special cases for null pointers in doPtrEq

chameco:
x86 code I've encountered in the wild sometimes contains the following pattern to check for a null pointer argument:

```asm
test %rdi, %rdi
jz .ret
```

Here `test` computes bitwise AND of the arguments and then sets flags. In order to set the zero flag, the semantics for `test` eventually `doPtrEq` the operand with 0. Right now, this always returns a fresh boolean variable, since one argument is a bitvector and the other is a pointer.


travitch (approved):
Good find - that makes sense.

PR #166, opened 2020-10-05
symbolic: Fix PackBits and UnpackBits

chameco:


travitch:
@joehendrix this change looks right to me, but I don't know the original context you had in mind when this was implemented. Are you aware of any other consumers of this code that we should test against?

andreistefanescu (approved):
lgtm

chameco:
I'm going to merge this, since I'm pretty sure it's correct and we need it for some time-sensitive work. If other changes are necessary, I'll be happy to make them in another PR.

PR #167, opened 2020-10-05
x86: Implement semantics for a few instructions

chameco:
Namely:
- `aesenc`, `aesenclast`, `aesdec`, `aesdeclast`, `aeskeygenassist`
- `pclmulqdq`
- `psrl{w,d,q,dq}`
- `pshufb`
- `shufps`
- `vpcmpgtd`

The AESNI instructions and `pclmulqdq` work in the same way as the existing `vaesenc`/`vaesenclast`/`vpclmulqdq`, using uninterpreted functions.

### DIFF
--- a/x86_symbolic/src/Data/Macaw/X86/Crucible.hs
+++ b/x86_symbolic/src/Data/Macaw/X86/Crucible.hs
@@ -241,6 +241,21 @@ evalApp' sym ev = C.evalApp (symIface sym) (symTys sym) logger evalExt ev
   evalExt :: fun -> EmptyExprExtension g a -> IO (RegValue sym a)
   evalExt _ y  = case y of {}
 
+pureSemAESNI
+  :: forall sym. IsSymInterface sym
+  => Sym sym
+  -> (SymFuns sym -> SymFn sym (EmptyCtx ::> BaseBVType 128 ::> BaseBVType 128) (BaseBVType 128))
+  -> (AtomWrapper (RegEntry sym) (M.BVType 128))
+  -> (AtomWrapper (RegEntry sym) (M.BVType 128))
+  -> IO (RegValue sym (ToCrucibleType (M.BVType 128)))
+pureSemAESNI sym proj x y =
+  do let f = proj (symFuns sym)
+         s = symIface sym
+     state <- toValBV s x
+     key <- toValBV s y
+     let ps = extend (extend empty state) key
+     llvmPointer_bv s =<< applySymFn s f ps
+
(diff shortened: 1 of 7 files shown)


x86_symbolic/src/Data/Macaw/X86/Crucible.hs
```diff
@@ -246,7 +276,16 @@ pureSem sym fn = do
...
+    M.PShufb sw (AtomWrapper vxs) (AtomWrapper vys) ->
+      case sw of
+        M.SIMDByteCount_XMM
+          | Just pxs <- V.fromList n16 $ DV.toList $ regValue vxs
+          , Just pys <- V.fromList n16 $ DV.toList $ regValue vys -> do
+              xs <- mapM (pure . ValBV n8 <=< projectLLVM_bv (symIface sym)) pxs
+              ys <- mapM (pure . ValBV n8 <=< projectLLVM_bv (symIface sym)) pys
+              let res = DV.fromList $ V.toList $ shuffleB xs ys
+              mapM (llvmPointer_bv (symIface sym) <=< evalE sym) res
+        _ -> error "PShufB"
```
travitch, line 288: Can we make this error message a bit more descriptive so that we'll know what to do when we see it?

x86_symbolic/src/Data/Macaw/X86/Crucible.hs
```diff
@@ -429,13 +468,38 @@ pureSem sym fn = do
...
+      do let f = fnClMul (symFuns sym)
+             s = symIface sym
+         src1 <- toValBV s x
+         src2 <- toValBV s y
+         let ps = extend (extend empty src1) src2
+         llvmPointer_bv s =<< applySymFn s f ps
+
+    M.AESNI_AESEnc x y -> pureSemAESNI sym fnAesEnc x y
+    M.AESNI_AESEncLast x y -> pureSemAESNI sym fnAesEncLast x y
+    M.AESNI_AESDec x y -> pureSemAESNI sym fnAesDecLast x y
```
travitch, line 481: Should this just be `fnAesDec`?

travitch (commented):
I did my best to review this, but I didn't look in detail at the semantics in the ISA manual.  There is just one comment that I think you might want to really look at

x86_symbolic/src/Data/Macaw/X86/Crucible.hs
chameco: Good catch - apparently I'm cursed to make these sorts of copy-paste errors twice over, as 0ff303d shows

PR #168, opened 2020-10-15
Init 8.6.5 testing

LisannaAtGalois:
Adds 8.6.5 and 8.10.1 to the ubuntu-latest GHC build matrix. The warnings as errors configuration is currently causing the 8.10.1 job to fail.

Is treating warnings as errors useful for CI? It causes compilation to stop after the first warning, preventing additional warnings from being displayed. Local testing indicates that 8.10.1 builds without errors if warnings are not treated as errors.


LisannaAtGalois:
@kquick should anyone else be added to this for reviewing?

travitch:
@joehendrix Will have the strongest preferences on this topic.

joehendrix:
@LisannaAtGalois Yes, I insist on warnings as errors for as many projects as willing to do it.  Otherwise warnings will get introduced and not fixed whereas I can always see all the warnings locally if needed.

joehendrix:
@LisannaAtGalois The additions look fine to me.  I think if you have the Haskell experience needed to fix the warnings, then that would be much appreciated. If not, I'd suggest droping 8.10.1 from the list of versions, then a separate followup could be made to fix the 8.10.1 warnings.

LisannaAtGalois:
> @LisannaAtGalois The additions look fine to me. I think if you have the Haskell experience needed to fix the warnings, then that would be much appreciated. If not, I'd suggest droping 8.10.1 from the list of versions, then a separate followup could be made to fix the 8.10.1 warnings.

@joehendrix I'll see how many I can get through, and I've got help for the ones that are still beyond me :)

LisannaAtGalois:
@joehendrix this should be ready to go in, but it looks like my changes to the job specification aren't matching up with the required status check configuration in the branch protection settings. I don't have admin on this repo, so I think you'll have to disable the check and then re-enable it with the new jobs selected after merging.

jared-w (approved):

kquick (approved):

joehendrix:
Should be good to go.

PR #172, opened 2020-10-21
Extract single Assignment element without redundant pattern case

kquick:
Resolves redundant case expression warnings under GHC 8.10.


travitch (approved):

joehendrix:
Looks fine to me.

PR #174, opened 2020-10-31
Support for mixed ARM/Thumb binaries

travitch:
It seems to work, but tests of a few more configurations are required


travitch:
This works for some test cases now. It comes with some minor caveats (e.g., the addresses of Thumb functions have the low bit set, which is technically not quite true, but predictable and somewhat reasonable - also very difficult to fix in a principled way).

kquick (approved):
Looks good!

PR #176, opened 2020-11-11
x86: Special case for sbb with duplicated operand

chameco:
This better handles cases like `sbb rax, rax`, where we know that the result will be `-cf` regardless of the value in `rax`.


travitch:
Is this behavior documented somewhere? I don't see it described in the ISA manual. Is the issue here potential overflow? Does it save us from reading the value from memory and let us produce a guaranteed simpler formula?

joehendrix:
I'd also be interested in knowing what this is trying to address.  I'd also prefer if you use the if-then-else style that is elsewhere used in `macaw` of
```
   if cond then
     trueVal
    else
     falseVal
```

chameco:
Thanks for the feedback! I think I've addressed the formatting - I'd like to keep things consistent with the rest of the codebase, so please let me know if there's anything else I can do to that end.

> Does it save us from reading the value from memory and let us produce a guaranteed simpler formula?

This is the intention; `sbb` computes `dst - (src + cf)`, and so e.g. `sbb rax, rax` is `rax - (rax + cf) = -cf`. This simplication is helpful especially when the operand register is uninitialized, which tends to complicate things down the line.

I don't know if this pattern is necessarily documented anywhere, but it seems to be a relatively common idiom. (At least, it's not too difficult to get recent Clang to emit it: https://godbolt.org/z/dqTTK6).

travitch:
Thanks. I'm happy with that rationale, but would really like it in the source next to the decision, since that is definitely non-obvious. It looks like the flags are computed independently so there shouldn't be any concern about missing flag values. Also, it looks like this could only fire for register operands (since the instruction can't have two memory operands), so we wouldn't miss any faults using this definition for verification.

chameco:
@travitch I've added a comment explaining the special case.

travitch (approved):

joehendrix:
This seems fine to me as is.

The other thought I has was that changing the logic for the generic case to `(dst - src) - cf` and adding a special  case to the definition of `.-` would have the side effect of simplifying any redundant subtractions and not just this one.
I don't know how common that is, but it'd move the simplifying logic to `Monad.hs` which is where most of the other simplification logic is.

joehendrix (approved):

PR #177, opened 2020-11-17
x86: Add semantics for aesimc

chameco:
This is the last of the non-AVX AESNI instructions. It follows the same pattern as the others (calling an uninterpreted function).


x86_symbolic/src/Data/Macaw/X86/Crucible.hs
```diff
@@ -213,6 +217,9 @@ newSymFuns s =
      fnAesDecLast <- bin "aesDecLast"
      fnAesKeyGenAssist <- bin "aesKeyGenAssist"
      fnClMul      <- bin "clMul"
+     fnAesIMC <- case userSymbol "aesIMC" of
+       Left _ -> fail "Invalid symbol name"
+       Right a -> freshTotalUninterpFn s a (extend empty knownRepr) knownRepr
```
travitch, line 222: Just one comment - would it make sense to just use `safeSymbol` here? I know it can't fail - it is always unfortunate to have these cases sitting around.

travitch (approved):
Just left one comment, but it doesn't necessarily need to be addressed

PR #180, opened 2020-12-06
Updates to PPC tests for new elf-edit API.

kquick:


travitch (approved):
Looks good to me - I'm not super worried about ppc32 just because there are really no differences between 64 and 32

PR #183, opened 2020-12-17
x86: Fix semantics for pinsrw, add semantics for pinsr{b,d,q}

chameco:
Fixes are as follows:
- Treat the offset as relative to the end of the destination bitvector instead of the beginning. (A `FIXME` comment mentioned that this might need to be changed, and some testing corroborated this.)
- Only use some instruction-specific number of low bits from the immediate operand. This was resulting in different behavior when the immediate was "too big", for example something like `pinsrd xmm1, eax, 5` should result in an offset of `5 & 0b11 = 1`.

### DIFF
--- a/x86/src/Data/Macaw/X86/Semantics.hs
+++ b/x86/src/Data/Macaw/X86/Semantics.hs
@@ -2451,11 +2451,11 @@ def_pselect mnem op sz = defBinaryLV mnem $ \l v -> do
 
 -- | PINSRW Insert word
 exec_pinsrx :: (1 <= w, 1 <= c)
-            => NatRepr w
-            -> NatRepr c
-            -> Location (Addr ids) (BVType c)
-            -> BVExpr ids w
-            -> Word8
+            => NatRepr w -- ^ Width of element to insert
+            -> NatRepr c -- ^ Width of destination registr
+            -> Location (Addr ids) (BVType c) -- ^ Destination operand
+            -> BVExpr ids w -- ^ Source operand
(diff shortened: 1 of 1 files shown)


x86/src/Data/Macaw/X86/Semantics.hs
```diff
@@ -2450,23 +2450,51 @@ def_pselect mnem op sz = defBinaryLV mnem $ \l v -> do
 -- PEXTRW Extract word
 
 -- | PINSRW Insert word
-exec_pinsrw :: Location (Addr ids) XMMType -> BVExpr ids 16 -> Word8 -> X86Generator st ids ()
-exec_pinsrw l v off = do
+exec_pinsrx :: (1 <= w, 1 <= c)
+            => NatRepr w
+            -> NatRepr c
```
travitch, line 2455: Could we get a comment explaining what `w` and `c` are here?

Are the mask values just taken from the ISA manual?

chameco: I've added some Haddock comments to those parameters.

The ISA manual specifies to use the 1, 2, 3, or 4 least significant bits of the immediate (depending on the instruction / whether the destination is MMX or XMM for `pinsrw`). I can add comments explaining the masks in this light if you'd like. I could also a binary literal which makes the masks more apparent, but this would require enabling the `BinaryLiterals` extension.

travitch (approved):
Thanks, that looks good to me.

PR #185, opened 2021-01-07
macaw-symbolic: record original block endings when making a CFG slice

danmatichuk:
Allow `mkBlockSliceCFG` to take global variable that it uses to record how the block originally exited before rewriting


symbolic/src/Data/Macaw/Symbolic.hs
```diff
@@ -892,6 +909,97 @@ toCoreCFG :: MacawSymbolicArchFunctions arch
...
+  | MacawBlockEndReturn
+  | MacawBlockEndBranch
+  | MacawBlockEndArch
+  | MacawBlockEndFail
+  deriving (Eq, Ord, Enum, Bounded)
+
+data MacawBlockEnd arch = MacawBlockEnd MacawBlockEndCase !(Maybe (M.ArchSegmentOff arch))
+
+-- we encode 'MacawBlockEndCase' as a 3-bit bitvector
+type MacawBlockEndType arch = C.StructType (Ctx.EmptyCtx Ctx.::> C.BVType 3 Ctx.::> C.MaybeType (MM.LLVMPointerType (M.ArchAddrWidth arch)))
```
travitch, line 926: Can we turn this into a haddock comment for nicer rendering, and can we document what the second struct element is?

symbolic/src/Data/Macaw/Symbolic.hs
```diff
@@ -892,6 +909,97 @@ toCoreCFG :: MacawSymbolicArchFunctions arch
...
+data MacawBlockEndCase =
+    MacawBlockEndJump
+  | MacawBlockEndCall
+  | MacawBlockEndReturn
+  | MacawBlockEndBranch
+  | MacawBlockEndArch
+  | MacawBlockEndFail
+  deriving (Eq, Ord, Enum, Bounded)
+
+data MacawBlockEnd arch = MacawBlockEnd MacawBlockEndCase !(Maybe (M.ArchSegmentOff arch))
```
travitch, line 923: What is the contained address referring to?

travitch (changes requested):

travitch (approved):
Thanks, that makes sense to me

PR #190, opened 2021-02-15
Update to follow changes in What4.  Nat is no longer a base type

robdockins:


travitch (changes requested):
It looks like the version bound on what4 in macaw-symbolic.cabal needs to be raised. Other than that this looks good

PR #192, opened 2021-03-03
symbolic: Some special-case handling for bitwise operations

chameco:
The special-cases in question are AND/XOR between a pointer and a bitvector when the alignment of the pointer's base is known and sufficiently large. We should be able to safely apply the bitwise operation to the pointer's offset if the operation only affects low bits that must be zero in the base.


travitch (approved):
This is great. You might be able to improve the performance of memory operations by taking a similar alignment approach for the memory read and write primitives, which currently assume byte alignment. Or that might be slower having to look up the allocation. It could also let you catch possible alignment errors that we would otherwise miss.

PR #193, opened 2021-03-19
Remove isValidPtr checks in doPtrSub and doPtrEq.

andreistefanescu:
The operations are correct even when the pointers are not valid LLVM
pointers.


travitch (approved):
I'm aligned with these changes. Just to add color to the PR description: while it is illegal to use invalid pointers, they are regularly constructed in various transient ways (usually due to compiler optimizations), and are harmless when created.

They are still checked for validity when they are dereferenced.

PR #196, opened 2021-03-24
x86: Add semantics for SHA256 instructions

chameco:


x86_symbolic/src/Data/Macaw/X86/Crucible.hs
```diff
@@ -495,6 +550,27 @@ pureSem sym fn = do
          let ps = extend empty src
          llvmPointer_bv s =<< applySymFn s f ps
 
+    M.SHA_sigma0 x -> pureSemSymUn sym fnShasigma0 x
+    M.SHA_sigma1 x -> pureSemSymUn sym fnShasigma1 x
+    M.SHA_Sigma0 x -> pureSemSymUn sym fnShaSigma0 x
+    M.SHA_Sigma1 x -> pureSemSymUn sym fnShaSigma1 x
+    M.SHA_Ch x y z ->
+      do let f = fnShaCh (symFuns sym)
```
andreistefanescu, line 558: nit: use `pureSemSymBin`

andreistefanescu (approved):

x86/src/Data/Macaw/X86/ArchTypes.hs
```diff
@@ -917,6 +952,12 @@ instance IsArchFn X86PrimFn where
       AESNI_AESDecLast x y -> sexprA "aesdeclast" [pp x, pp y]
       AESNI_AESKeyGenAssist x i -> sexprA "aeskeygenassist" [pp x, ppShow i]
       AESNI_AESIMC x -> sexprA "aesimc" [pp x]
+      SHA_sigma0 x -> sexprA "sigma0" [pp x]
+      SHA_sigma1 x -> sexprA "sigma1" [pp x]
+      SHA_Sigma0 x -> sexprA "Sigma0" [pp x]
+      SHA_Sigma1 x -> sexprA "Sigma1" [pp x]
+      SHA_Ch x y z -> sexprA "Ch" [pp x, pp y, pp z]
+      SHA_Maj x y z -> sexprA "Maj" [pp x, pp y, pp z]
```
travitch, line 960: Could we prefix these names with `sha_` or something similar? If printed without context, they would look pretty mysterious

x86/src/Data/Macaw/X86/ArchTypes.hs
```diff
@@ -738,6 +738,29 @@ data X86PrimFn f tp where
...
+  SHA_Ch
+    :: !(f (BVType 32))
+    -> !(f (BVType 32))
+    -> !(f (BVType 32))
+    -> X86PrimFn f (BVType 32)
+  SHA_Maj
+    :: !(f (BVType 32))
+    -> !(f (BVType 32))
+    -> !(f (BVType 32))
+    -> X86PrimFn f (BVType 32)
```
travitch, line 762: I'm not super thrilled with the short names, but I understand why they are that way (and that there aren't really better names). My only request is a pointer here to what these are.  Something like: These represent operations in the SHA256 instructions, see SHA256RNDS2, SHA256MSG1, and SHA256MSG2 in the ISA manual.

I also found concise definitions of `Ch`, `Maj`, and the sigma functions (https://web.archive.org/web/20130526224224/http://csrc.nist.gov/groups/STM/cavp/documents/shs/sha256-384-512.pdf on page 8) that might be nice to at least reference.

travitch (approved):
Okay, I see where these operation names come from. I'm not thrilled with the short names, and would appreciate the requested comments. Otherwise I approve

PR #197, opened 2021-03-24
New vector operations; CI updates

joehendrix:


LisannaAtGalois (commented):
Thanks for keeping me in the loop on this stuff even if I can't always look at it the same day ^^

How are the freeze files for the macaw repo generated? Is it plain invocation of cabal freeze, or are there special considerations that are manually maintained, like what we do for the [saw-script repo](https://github.com/GaloisInc/saw-script#notes-on-freeze-files)?

joehendrix:
I just use a plain `cabal freeze` and manually update.  Cross platform freeze files has not been a priority

PR #202, opened 2021-04-05
[symbolic] Rename EvalStmtFunc to MacawEvalStmtFunc

kquick:
There is already an `EvalStmtFunc` defined by Crucible, and while the two are related, the type signature is different.  Changing the name may cause some downstream impact, but there is no other change than the name (other than some formatting updates) and this helps disambiguate any references.


travitch (approved):
Here is me hoping that there aren't too many downstream things depending on this. It looks like just pate, and it isn't hard to update.

kquick:
That's my hope as well.  In my searching, the Crucible one looks broadly used but hardly any uses of the Macaw version, so the renaming impact should be low. 🤞

PR #204, opened 2021-04-13
WIP: Introduce incremental computation monad; refactoring of app output.

joehendrix:


joehendrix:
I've updated the name and merging this as reopt uses the incremental computation capability now, and I want to do some more work before getting back to JSON output.

PR #216, opened 2021-07-12
x86: Fix semantics for BSF and BSR instructions

chameco:
The old implementations weren't matching what I observed on my machine. I think old `BSR` computed `BSF`, while the old `BSF` computed `w - BSR` (maybe also off by one).

I've switched things around so that the results match hardware for the few inputs that I've tried. I'd love to include a test - if there's a good place for those, I'd appreciate a pointer in the right direction.


travitch:
There is a test suite for macaw-x86-symbolic. It loads an executable (with symbol information) and symbolically executes each function it finds (function calls within test cases are not currently supported). The test harness then asserts that the return value of each test function is non-zero.

travitch (approved):

PR #224, opened 2021-08-23
Adapt to GaloisInc/crucible#794

RyanGlScott:
GaloisInc/crucible#794 increases the number of functions that use implicit `MemOptions`, including a handful of key LLVM memory model–related functions. As a result, many parts of `macaw` need to add implicit `?memOpts` parameters to accommodate to this change.


travitch (approved):
Straightforward, thanks. I lament the additional constraints, but will survive it.

PR #230, opened 2021-09-10
Add support for FSBASE and GSBASE registers on x86_64

bboston7:
This change adds support for the FSBASE and GSBASE registers on x86_64.
It also lays the groundwork for supporting additional special segment
registers on other architectures.  It does this by adding a callback
`LookupSegmentBasePointer` that takes an `ArchFn` indicating which
segment register is being read, and returns a base pointer to that
memory segment.


travitch (changes requested):
Can we make this just an argument to `x86_64MacawEvalFn`? The second most common architecture that actually supports segmentation is the AS/400, so I'm inclined to keep details like that out of the core libraries.

That would also make it easier for us to provide a default implementation (i.e., specialized to x86_64).

travitch (approved):
I like it. Thanks. When you merge, can you squash and add a note in the commit message about how to update client code?

PR #232, opened 2021-09-27
Add Show{,F} instances for MacawCrucibleValue

RyanGlScott:
I found these useful while printf debugging.


travitch (approved):

RyanGlScott:
The CI builders are down, so unfortunately we do not have CI confirmation that this change validates. I have built this locally, and since the patch is pretty small, I'm going to be bold and just land this.

PR #233, opened 2021-09-27
`stripPLTRead`: Avoid referencing dropped IDs in `ArchState`s

RyanGlScott:
It is possible for later `ArchState` updates to reference the `AssignId` of the `AssignStmt` that is dropped as a part of `stripPLTRead`, so make sure to prune such updates from the `ArchState` to avoid referencing the now out-of-scope `AssignId`.


base/src/Data/Macaw/Discovery.hs
```diff
@@ -898,6 +900,22 @@ stripPLTRead readId next rest =
...
+    -- It is possible for later ArchState updates to reference the AssignId of
+    -- the AssignStmt that is dropped, so make sure to prune such updates to
+    -- avoid referencing the now out-of-scope AssignId.
+    dropRefsTo :: AssignId ids tp -> Stmt arch ids -> Stmt arch ids
+    dropRefsTo stmtId stmt =
+      case stmt of
+        ArchState addr updates ->
+          ArchState addr $
+          MapF.filter (\v -> Some stmtId `Set.notMember` refsInValue v) updates
+        AssignStmt{}       -> stmt
```
RyanGlScott, line 913: I'm not sure if we need to worry about stale references appearing in these other forms of `Stmt`. (More generally, I'm not sure of the best way to test these changes.)

travitch: `Comment` and `InstructionStart` are fine because they don't carry any values with them.

`stripPLTRead` bails out if it finds any of the others, so it should be fine (i.e., you shouldn't see any relevant things in `dropRefsTo`)

travitch (approved):

base/src/Data/Macaw/Discovery.hs
RyanGlScott: Ah, right. I'll add some comments to this effect.

As far as testing goes, the test suite for `macaw-x86` seems to the best fit, as it has a collection of binaries that we could add to. But we'd likely need to fix #234 before we could successfully analyze such a binary.

RyanGlScott:
The CI builders are down, so unfortunately we do not have CI confirmation that this change validates. I have checked that all of the test suites pass locally with GHC 8.10, so I'll be bold and land this.

PR #243, opened 2021-11-12
AArch32: Support conditional returns

travitch:
Use the recently exposed parameterized block classification infrastructure to implement conditional returns in ARM as an architecture extension


macaw-aarch32/src/Data/Macaw/ARM/Identify.hs
```diff
@@ -74,3 +90,122 @@ isReturnValue absProcState val =
...
+  -> MA.AbsProcessorState (MC.ArchReg ARM.AArch32) ids
+  -> MC.Value ARM.AArch32 ids (MT.BVType (MC.ArchAddrWidth ARM.AArch32))
+  -> MC.Value ARM.AArch32 ids (MT.BVType (MC.ArchAddrWidth ARM.AArch32))
+  -> Maybe (MC.ArchSegmentOff ARM.AArch32)
+asReturnAddrAndConstant mem absProcState mRet mConstant = do
+  MA.ReturnAddr <- return (MA.transferValue absProcState mRet)
+  memAddr <- MC.valueAsMemAddr mConstant
+  segOff <- MC.asSegmentOff mem memAddr
+  when (not (isExecutableSegOff segOff)) $ do
+    fail ("Conditional return successor is not executable: " ++ show memAddr)
```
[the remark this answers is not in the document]
kquick: On further review of the use below, should this just return `Nothing` without bringing `fail` into the picture?

Lines 107-108 could become simply:
```
  guard (isExecutableSegOff segOff)
```

macaw-aarch32/tests/ARMTests.hs
```diff
@@ -113,17 +127,32 @@ getExpected expectedFilename = do
           ignoredBlocks = S.fromList (ignoreBlocks er)
       in return (expectedEntries, ignoredBlocks)
 
+escapeSlash :: Char -> Char
+escapeSlash '/' = '_'
+escapeSlash c = c
+
+toSavedMacawPath :: String -> FilePath
+toSavedMacawPath testName = "/tmp" </> name <.> "macaw"
```
kquick, line 135: I'd recommend simply writing to the current directory.  If this flag was ever enabled for CI I don't know that `/tmp` would necessarily be available, or if it was, there's a risk of the wrong output due to parallel tests.  You could also use the contents of a `TEST_OUTDIR` env var lookup with fallback to `.` if you wanted flexibility.

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -204,10 +206,18 @@ data MacawSymbolicArchFunctions arch
                       -> CrucGen arch ids s ())
      -- ^ Generate crucible for architecture-specific statement.
   , crucGenArchTermStmt :: !(forall ids s
-                               . M.ArchTermStmt arch ids
+                               . M.ArchTermStmt arch (M.Value arch ids)
+                               -- ^ The arch-specific terminator
```
kquick, line 210: Here also, document what the `(M.Value arch ids)` represents.

macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
```diff
@@ -96,26 +96,59 @@ data ARMBlockPrecond =
...
+  --
+  -- Note that it is unfortunate that we need this in addition to 'ReturnIf'. At
+  -- the part of the analysis where we are able to generate these block
+  -- terminators, we cannot generate new statements since we do not have the
+  -- right nonce generator. As a result, if we are returning when the condition
+  -- is false, we are not able to generate a negation (which would require
+  -- wrapping the original condition in an Assignment, which has an AssignId,
+  -- which requires a nonce). We work around this by just having an additional
+  -- block terminator.
+  ReturnIfNot :: f MT.BoolType -> ARMTermStmt f
```
kquick, line 114: If you combined these to `ReturnCond :: Bool -> f MT.BoolType -> ARMTermStmt f` (or even better, `ReturnCond :: ReturnsOnBranch -> f MT.BoolType -> ARMTermStmt f` if you moved `ReturnsOnBranch` to be visible here) then you'd only have a single constructor, which would save a lot of code duplication.  It doesn't look like the different constructors provides any pattern-matching value other than in `aarch32GenTermStmt`, which could just as easily match on `ReturnsOnBranch` constructors.

macaw-aarch32/src/Data/Macaw/ARM/Identify.hs
```diff
@@ -74,3 +90,122 @@ isReturnValue absProcState val =
...
+--
+-- Note that there are two cases below that could be handled. It seems unlikely
+-- that these would be produced in practice, so they are unhandled for now.
+conditionalReturnClassifier :: MAI.BlockClassifier ARM.AArch32 ids
+conditionalReturnClassifier = do
+  stmts <- CMR.asks MAI.classifierStmts
+  mem <- CMR.asks (MAI.pctxMemory . MAI.classifierParseContext)
+  regs <- CMR.asks MAI.classifierFinalRegState
+  absState <- CMR.asks MAI.classifierAbsState
+  Just (cond, nextIP, returnBranch, stmts') <- return (identifyConditionalReturn mem stmts regs absState)
```
kquick, line 187: Should the `fail` path here invoke (i.e. fall through to) the `defaultClassifier`?

macaw-aarch32/src/Data/Macaw/ARM/Identify.hs
```diff
@@ -74,3 +90,122 @@ isReturnValue absProcState val =
...
+                              )
+      return Parsed.ParsedContents { Parsed.parsedNonterm = F.toList stmts'
+                                   , Parsed.parsedTerm = Parsed.ParsedArchTermStmt term regs (Just nextIP)
+                                   , Parsed.intraJumpTargets = [fallthroughTarget]
+                                   , Parsed.newFunctionAddrs = []
+                                   , Parsed.writtenCodeAddrs = writtenAddrs
+                                   }
+    Jmp.TrueFeasibleBranch _ -> fail "Infeasible false branch"
+    Jmp.FalseFeasibleBranch _ -> fail "Infeasible true branch"
+    Jmp.InfeasibleBranch -> fail "Branch targets are both infeasible"
```
kquick, line 211: Rather than these being a `ClassifyFailure`, shouldn't these be `panic` to match the comment "It seems unlikely... so they are unhandled for now."?

kquick (approved):
I had some suggestions, but most are minor other than some of the `fail` condition handling.

macaw-aarch32/src/Data/Macaw/ARM/Identify.hs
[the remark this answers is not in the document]
travitch: Ah I think that used to be inside of the `Classify` monad, which collected those. I should probably restructure some more of this code to be inside of that monad to preserve the messages.

macaw-aarch32/src/Data/Macaw/ARM/Identify.hs
travitch: Yeah a pattern match failure will fall through (and its error message will be collected). I'll try to restructure so that the more precise messages are preserved.

base/src/Data/Macaw/CFG/AssignRhs.hs
```diff
@@ -76,14 +76,9 @@ type family ArchStmt (arch :: Kind.Type) = (stmt :: (Type -> Kind.Type) -> Kind.
...
--- The second type parameter is the ids phantom type used to provide
--- uniqueness of Nonce values that identify assignments.
---
 -- An architecture-specific terminal statement may have side effects and change register
--- values, it may or may not return to the current function.  If it does return to the
--- current function, it is assumed to be at most one location, and the block-translator
--- must provide that value at translation time.
-type family ArchTermStmt (arch :: Kind.Type) :: Kind.Type -> Kind.Type
+-- values, it may or may not return to the current function.
+type family ArchTermStmt (arch :: Kind.Type) :: (Type -> Kind.Type) -> Kind.Type
```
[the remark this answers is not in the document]
RyanGlScott: Oof, I didn't realize that `macaw` had its own, completely distinct data type named `Type`.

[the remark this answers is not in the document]
travitch: It sure is not ideal

macaw-aarch32/src/Data/Macaw/ARM/Identify.hs
travitch: I think I'd rather them turn into classify failures rather than hard-stop panics. A user could potentially provide a binary that looked like that, though I'm not concretely sure how.

PR #246, opened 2021-11-24
Revise handling of syscalls in AArch32 to match X86

travitch:
The old formulation (with system calls as block terminators) proved to be
impossible to implement properly. Handlers for syntax overrides have very
limited types (`IO`, rather than `OverrideSim`), which made symbolic branching
and reusing overrides impossible.

This change replaces the system call block terminator with an arch-specific
function that is translated into a function handle lookup (which is then
dispatched to with a call).

Unfortunately, this refactoring required combining the AArch32 simplification
module with the architecture extension definitions, due to the new translation
relying on the simplifier instance.


macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
```diff
@@ -137,25 +133,60 @@ instance TF.FunctorF ARMTermStmt where
...
+  ARMSyscall :: WI.W 24
+             -> f (MT.BVType 32) -- r0
+             -> f (MT.BVType 32) -- r1
+             -> f (MT.BVType 32) -- r2
+             -> f (MT.BVType 32) -- r3
+             -> f (MT.BVType 32) -- r4
+             -> f (MT.BVType 32) -- r5
+             -> f (MT.BVType 32) -- r6
+             -> f (MT.BVType 32) -- r7 (syscall number)
+             -> ARMPrimFn f (MT.TupleType [MT.BVType 32, MT.BVType 32])
```
RyanGlScott, line 176: Since you have comments stating which registers each of the arguments correspond to, you might as well add comments for the return registers as well. Similarly for `ThumbSyscall`'s return registers.

macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
```diff
@@ -137,25 +133,60 @@ instance TF.FunctorF ARMTermStmt where
...
+  -- account of the translation strategy.
+  --
+  -- The 'WI.W' is the immediate operand embedded in the opcode
+  --
+  -- The rest of the arguments are all of the registers that participate in the
+  -- syscall protocol (at least on Linux). Arguments are passed in r0-r6, while
+  -- the syscall number is in r7.
+  --
+  -- The system call can return up to two values (in r0 and r1)
+  ARMSyscall :: WI.W 24
```
RyanGlScott, line 167: Aside from the immediate operand, the types of `ARMSyscall` and `ThumbSyscall` are identical. Would it be cleaner to combine the two data constructors into one and have a separate data type for the immediate operand? Perhaps something like:

```hs
data ARMPrimFn (f :: MT.Type -> Type) tp where
  Syscall :: SyscallImm
          -> f (MT.BVType 32)
          -> f (MT.BVType 32)
          -> f (MT.BVType 32)
          -> f (MT.BVType 32)
          -> f (MT.BVType 32)
          -> f (MT.BVType 32)
          -> f (MT.BVType 32)
          -> f (MT.BVType 32)
          -> ARMPrimFn f (MT.TupleType [MT.BVType 32, MT.BVType 32])
  ...

data SyscallImm
  = ARMImm (WI.W 24)
  | ThumbImm (WI.W 8)
```

This would help avoid some noticeable code duplication involving the `*Syscall` data constructors (e.g., in `aarch32GenFn`). I'm not sure if this would work out to be less code overall, since there are other parts of the code that do different things for ARM and Thumb, but the idea seems tantalizing.

RyanGlScott (approved):

macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
travitch: Yeah this was bugging me too. I think I might just zero extend the immediate and use the same constructor for both. The kernel can't actually tell which instruction got you there.

PR #247, opened 2021-11-29
Add a concrete syntax for macaw

travitch:
This will enable caching analysis results in a text format that can be parsed
much more efficiently than re-running the entire fixed point analysis.

### DIFF
--- a/base/src/Data/Macaw/Syntax/Atom.hs
+++ b/base/src/Data/Macaw/Syntax/Atom.hs
@@ -1,0 +1,131 @@
+{-# LANGUAGE OverloadedStrings #-}
+-- | The atoms of the concrete syntax for macaw
+module Data.Macaw.Syntax.Atom (
+    Keyword(..)
+  , keywords
+  , AtomName(..)
+  , Atom(..)
+  )
+  where
+
+import qualified Data.Map.Strict as Map
+import qualified Data.Text as T
+import           Numeric.Natural ( Natural )
+
+-- | Macaw syntax keywords
+--
+-- These are the keywords for the *base* macaw IR (i.e., without
+-- architecture-specific extensions).  The architecture-specific operations are
+-- parsed as 'AtomName's initially and resolved by architecture-specific parsers
+-- at the atom level.
+data Keyword = BVAdd
+             | BVSub
+             | BVMul
+             | BVAdc
+             | BVSbb
+             | BVAnd
+             | BVOr
+             | BVXor
+             | BVShl
+             | BVShr
+             | BVSar
+             | PopCount
+             | Bsf
+             | Bsr
+             | BVComplement
+             | Mux
+             | Lt
+             | Le
+             | Sle
+             | Slt
+             -- Syntax
+             | Assign
+             -- Statements
+             | Comment
+             | InstructionStart
+             | WriteMemory
+             | CondWriteMemory
+             -- Expressions
+             | ReadMemory
+             -- Boolean operations
+             | Eq_
+             | Not_
+             | And_
+             | Or_
+             | Xor_
+             -- Endianness
+             | BigEndian
+             | LittleEndian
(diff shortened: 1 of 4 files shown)


base/src/Data/Macaw/Syntax/Parser.hs
```diff
@@ -0,0 +1,269 @@
...
+      SomeTyped _tp app <- asApp a
+      return (Some (MC.EvalApp app))
+
+-- | Forms:
+--
+--  * @(comment "msg")@
+--  * @(instruction-start addr decoded-asm-text)@
+--  * @(write-memory addr mem-rep value)@
+--  * @(cond-write-memory cond addr mem-rep value)@
+--  * @(reg := rhs)@
```
RyanGlScott, line 194: The Lisper in me is somewhat weirded out by the fact that this is `(reg := rhs)` and not `(:= reg rhs)`, especially since this is (AFAICT) the only operator that is given infix treatment.

base/src/Data/Macaw/Syntax/SExpr.hs
```diff
@@ -0,0 +1,231 @@
...
+
+deriving instance Eq MacawSyntaxError
+deriving instance Ord MacawSyntaxError
+
+data ParserContext arch ids =
+  ParserContext { memory :: MM.Memory (MC.ArchAddrWidth arch)
+                , nonceGen :: PN.NonceGenerator (ST ids) ids
+                }
+
+data ParserState arch ids =
```
RyanGlScott, line 121:
```suggestion
newtype ParserState arch ids =
```

base/src/Data/Macaw/Syntax/Atom.hs
```diff
@@ -0,0 +1,108 @@
...
+newtype AtomName = AtomText T.Text
+  deriving (Eq, Ord, Show)
+
+data Atom = Keyword !Keyword -- ^ Keywords include all of the built-in expressions and operators
+          | AtomName !AtomName -- ^ Non-keyword syntax atoms (to be interpreted at parse time)
+          | Register !Natural  -- ^ A numbered local register (e.g., @r12@)
+          | AddressWord !Natural -- ^ An arbitrary address rendered in hex ('ArchAddrWord')
+          | SegmentOffset !Natural -- ^ A segment offset address rendered in hex (validation against the Memory object is required)
+          | Integer_ !Integer    -- ^ Literal integers
+          | Natural_ !Natural    -- ^ Literal naturals
```
RyanGlScott, line 72: What exactly is the difference between a `Natural` and an `Integer`? Is an `Integer` just a `Natural` with a ± sign? I ask since there are some operations that seem to require an `Integer` rather than a `Natural` argument, such as `bv`, whose second argument must be an `Integer`. Does this imply that `bv 8 +0` is legal but `bv 8 0` is illegal?

travitch: I haven't decided yet. There are a bunch of places where a negative number is not legal but where I don't want to have partial cases due to range checks on `Integer`. My current thought is that I would require a sign on `Integer` (even for positive numbers), making all unsigned values `Natural`.

I think your observation about `bv` is good (and pointing out an error in the code): I would prefer that it be `(bv 8 0)` (i.e., the actual value is also a `Natural`).

Note that this entire syntax is not really meant for human consumption at all. This will all be machine generated and machine parsed. It is only textual so that I don't go crazy debugging it, so I don't mind if it is ugly.

base/src/Data/Macaw/Syntax/SExpr.hs
travitch: I'll do that if the parser state doesn't gain any additional fields by the time I'm done with the implementation

langston-barrett:
Closing, as it's been a few years and we haven't had an urgent need for this.

PR #257, opened 2022-01-11
Update with changes flowing from GaloicInc/crucible#945.

robdockins:
This mostly deals with the splitting of the old `sym` type into
two: one for dealing with expression creation, and a new simulator
backend type for dealing with control-flow and assertions.

### DIFF
--- a/macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs
+++ b/macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs
@@ -75,10 +75,11 @@ stmtSemantics :: (C.IsSymInterface sym, 1 <= SP.AddrWidth v)
               -> S v sym rtp bs r ctx
               -> IO (C.RegValue sym C.UnitType, S v sym rtp bs r ctx)
 stmtSemantics _sf stmt s =
+  C.withBackend (s ^. C.stateContext) $ \bak ->
   case stmt of
     MP.Attn -> do
       let reason = C.GenericSimError "ppc_attn"
-      C.addFailedAssertion (s ^. C.stateSymInterface) reason
+      C.addFailedAssertion bak reason
     -- These are cache hints that can't be observed in our current memory model
     -- (i.e., they require concurrency to be observed).
     --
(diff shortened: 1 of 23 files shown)


macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs
```diff
@@ -117,69 +118,69 @@ funcSemantics :: (C.IsSymInterface sym, MS.ToCrucibleType mt ~ t, 1 <= SP.AddrWi
...
-      let sym = s ^. C.stateSymInterface
-      lhs' <- toValBV sym lhs
-      rhs' <- toValBV sym rhs
+      lhs' <- toValBV bak lhs
+      rhs' <- toValBV bak rhs
       -- FIXME: Make sure that the semantics when rhs == 0 exactly match PowerPC.
       v <- LL.llvmPointer_bv sym =<< C.bvSdiv sym lhs' rhs'
       return (v, s)
-    MP.FPNeg (_ :: MT.FloatInfoRepr fi) x -> withSymFPUnOp s x $ \sym x' ->
+    MP.FPNeg (_ :: MT.FloatInfoRepr fi) x -> withSymFPUnOp s x $ \_sym x' ->
```
robdockins, line 136: This ends up being kind of a strange pattern.  I left a bunch of `_sym` arguments in place for now to persue smaller diffs, but maybe it's worth revisiting.

symbolic/src/Data/Macaw/Symbolic/MemOps.hs
```diff
@@ -154,12 +151,16 @@ ptrBase = fst . llvmPointerView
 -- The 'GlobalMap' is responsible for 1) determining which LLVM allocation
 -- contains the relocatable value, and 2) returning the corresponding address in
 -- that allocation.
-type GlobalMap sym mem w
-  = sym                        {-^ The symbolic backend -} ->
-    GetIntrinsic sym mem       {-^ The global handle to the memory model -} ->
-    RegValue sym NatType       {-^ The region index of the pointer being translated -} ->
-    RegValue sym (BVType w)    {-^ The offset of the pointer into the region -} ->
-    IO (LLVMPtr sym w)
+newtype GlobalMap sym mem w =
```
RyanGlScott, line 154: This becoming a newtype would be worth mentioning in the changelog.

x86_symbolic/src/Data/Macaw/X86/Crucible.hs
```diff
@@ -863,31 +877,43 @@ unpack2 ::
...
   IO (E sym (BVType w))
-getBitVal sym (AtomWrapper x) =
-  do v <- projectLLVM_bv sym (regValue x)
+getBitVal bak (AtomWrapper x) =
+  do v <- projectLLVM_bv bak (regValue x)
      case regType x of
        LLVMPointerRepr w -> return (ValBV w v)
        _ -> error "getBitVal: impossible"
 
+getBitVal' ::
```
RyanGlScott, line 899: We (for some definition of "we") should leave a comment explaining why both `getBitVal` and `getBitVal'` exist and why you'd want to use the latter over the former in one particular spot. (I know that I'm at fault here for offering a patch that introduced `getBitVal'`, but I'd hate to see this become a permanent fixture without some kind of accompanying documentation.)

robdockins: I was wondering if this should get pulled out into a separate PR that lands before this one.  I think there are still some questions to answer about if this was the right way to fix this.

macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs
travitch: I think that is fine. Next major update we need to make to the semantics, we can revisit

x86_symbolic/src/Data/Macaw/X86/Crucible.hs
travitch: I'll work on a separate PR for this

travitch (approved):

PR #261, opened 2022-01-21
x86: Fix failing proof obligations due to EvenParity

### LINKED ISSUE #260: Macaw update breaks some tests
While attempting to bump the Macaw submodule, I discovered some changes that caused breakages.  I don't fully appreciate the implications, but bisection identifies the first bad commit as: https://github.com/GaloisInc/macaw/tree/9c426986ff5303530d3550b17abd5b03411d9751

This breaks the `test_bitfield_x86` integration test, and I think is also implicated in problems with the s2n/sike proof.

travitch:
See the writeup in Crucible.hs in this commit for details. In short, the recent
changes to generalize `PtrAdd` triggered a failing proof obligation due to a use
of `llvmPointer_bv`.  The new implementation is as sound as the previous one,
but more general.

Fixes #260


symbolic/src/Data/Macaw/Symbolic/Testing.hs
```diff
@@ -198,6 +238,13 @@ functionName dfi = maybe addrName fromByteString (MD.discoveredFunSymbol dfi)
 -- *translation* of programs into Crucible, while 'MS.MacawArchEvalFn' is used
 -- during symbolic execution (to provide interpretations to
 -- architecture-specific primitives).
+--
+-- In addition to symbolically executing the function to produce a Sat/Unsat
+-- result, this function will attempt to verify all generate side conditions if
+-- the name of the function being simulated begins with @test_and_verify_@ (as
+-- opposed to just @test@).  This is necessary for testing aspects of the
```
RyanGlScott, line 245: It would be good to make a note of this convention in the `README`s in each of the test suites' top-level directories (`macaw-aarch32-symbolic/tests/README.org`, `macaw-ppc-symbolic/tests/README.org`, and `x86_symbolic/tests/README.org`).

x86_symbolic/src/Data/Macaw/X86/Crucible.hs
```diff
@@ -869,6 +872,11 @@ unpack2 sym e w c v1 v2 k =
        v2' <- getBitVal sym v2
        k (V.fromBV e n c v1') (V.fromBV e n c v2')
 
+-- | Extract the offset from an LLVMPointer
+--
+-- Note that this version generates an /assertion/ that the block number is
+-- zero, and should be used carefully.
```
RyanGlScott, line 878: It would be worth mentioning `getPointerOffset` here as an alternative that doesn't generate an assertion.

RyanGlScott (approved):

Issue #234, opened 2021-09-28
Support binaries that invoke functions from shared libraries


RyanGlScott:
We would like to have `macaw` be able to support binaries that invoke functions from shared libraries. This would (but is not limited to) doing the following:

1. Updating this case in `addMacawParsedTermStmt` in `macaw`:

    https://github.com/GaloisInc/macaw/blob/a336895da7158dfc0ad22975ac7efc0f43dbcbe6/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs#L1486-L1487
2. Update `LookupFunctionHandle` in `macaw-symbolic` to have some understanding of PLT entries (see https://refspecs.linuxfoundation.org/ELF/zSeries/lzsabi0_zSeries/x2251.html), as calls to shared library functions are done indirectly though the PLT.

PR #262, opened 2022-02-02
Add support for standalone PIEs

bboston7:
This changeset adds support for standalone position independent
executables (PIEs) that do not make use of procedure linkage tables.  It
does so by adding relative address support to `populateSegmentChunk` and
adding an additional simplification rule for Aarch32.

This covers part of the work for #234.

### DIFF
--- a/macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
+++ b/macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
@@ -19,7 +19,7 @@
 module Data.Macaw.ARM.Arch where
 
 import           Control.Applicative ( (<|>) )
-import           Data.Bits ( (.&.), shiftR )
+import           Data.Bits ( (.&.), shiftL, shiftR )
 import qualified Data.BitVector.Sized as BVS
 import           Data.Kind ( Type )
 import qualified Data.Macaw.ARM.ARMReg as ARMReg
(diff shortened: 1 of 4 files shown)


macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
```diff
@@ -908,3 +909,58 @@ negatedMux app = do
...
+        Just PC.Refl -> do
+          -- NOTE: Performing the shift and multiplication operations on the
+          -- offset of a relative address like this is only correct when the
+          -- base address is 4-byte aligned.  However, we expect this to always
+          -- be true because in practice relocatable addresses are used with
+          -- ALSR which must preserve page alignment at a larger power of 2 (at
+          -- least 4KB on Linux).  Thus, this simplification is safe for small
+          -- shift/multiply pairs (like 2 and 4 here) but shound not be
+          -- expanded to cover the general case.
+          let resOffset = (offset `shiftR` 2) * 4
```
bboston7, line 964: @travitch I want to specifically call this part out.  Does this assumption about offsets being 4 byte aligned seem reasonable to you?

travitch: I'm trying to think of a case where we could get into trouble. The cases I came up with would involve dynamic allocations, which would not apply here. I believe this is fine with this documentation here.

travitch (commented):
I'm broadly aligned, but see my question on the `NoOp` operation

macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
```diff
@@ -908,3 +909,58 @@ negatedMux app = do
...
+  case MT.typeRepr val of
+    MT.BVTypeRepr valWidth ->
+      case PC.testEquality valWidth targetSize of
+        Nothing -> Nothing
+        Just PC.Refl -> do
+          -- NOTE: Performing the shift and multiplication operations on the
+          -- offset of a relative address like this is only correct when the
+          -- base address is 4-byte aligned.  However, we expect this to always
+          -- be true because in practice relocatable addresses are used with
+          -- ALSR which must preserve page alignment at a larger power of 2 (at
```
RyanGlScott, line 960:
```suggestion
          -- ASLR which must preserve page alignment at a larger power of 2 (at
```

macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
```diff
@@ -908,3 +909,58 @@ negatedMux app = do
...
+          -- least 4KB on Linux).  Thus, this simplification is safe for small
+          -- shift/multiply pairs (like 2 and 4 here) but shound not be
```
RyanGlScott, line 962:
```suggestion
          -- shift/multiply pairs (like 2 and 4 here) but should not be
```

macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
```diff
@@ -908,3 +909,58 @@ negatedMux app = do
...
+          -- expanded to cover the general case.
+          let resOffset = (offset `shiftR` 2) * 4
+          let res = MC.RelocatableValue w (MC.MemAddr base resOffset)
+          return (MC.NoOp (MT.BVTypeRepr valWidth) res)
```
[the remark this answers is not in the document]
bboston7: Yep, this worked.

travitch (approved):

PR #263, opened 2022-02-18
Translate PLTStubs as tail calls

### LINKED ISSUE #234: Support binaries that invoke functions from shared libraries

bboston7:
Closes #234 

This change modifies `addMacawParsedTermStmt` to translate `PLTStub`s as
tail calls.

### DIFF
--- a/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
@@ -1575,10 +1575,9 @@ addMacawParsedTermStmt blockLabelMap externalResolutions thisAddr tstmt = do
       -- have more flexibility when working with shared libraries.
       unless (MapF.null updatedRegs) (error "Do not support updated regvals")
       curRegs <- getRegs
-      let tps = typeCtxToCrucible $ fmapFC M.typeRepr $ crucGenRegAssignment archFns
       fh <- evalMacawStmt (MacawLookupFunctionHandle (crucArchRegTypes archFns) curRegs)
-      newRegs <- evalAtom $ CR.Call fh (Ctx.singleton curRegs) (C.StructRepr tps)
-      addTermStmt $ CR.Return newRegs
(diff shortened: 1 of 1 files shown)


symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -1567,8 +1567,18 @@ addMacawParsedTermStmt blockLabelMap externalResolutions thisAddr tstmt = do
...
+      -- register state through a 'MacawLookupFunctionHandle'.  If the need
+      -- arises we may want to add a new 'MacawStmtExtension' (something like
+      -- MacawLookupPLTStubHandle) containing '_target' to allow macaw users to
+      -- have more flexibility when working with shared libraries.
+      unless (MapF.null updatedRegs) (error "Do not support updated regvals")
+      curRegs <- getRegs
+      let tps = typeCtxToCrucible $ fmapFC M.typeRepr $ crucGenRegAssignment archFns
+      fh <- evalMacawStmt (MacawLookupFunctionHandle (crucArchRegTypes archFns) curRegs)
+      newRegs <- evalAtom $ CR.Call fh (Ctx.singleton curRegs) (C.StructRepr tps)
+      addTermStmt $ CR.Return newRegs
```
travitch, line 1581: Could these be combined into a `CR.TailCall` terminator?

PR #203, opened 2021-04-08
Only fix personality to (MS.MacawSimulatorState sym) where required.

kquick:
In preparation for Polyglot support, this changes various locations that had unnecessarily hard-coded the personality to `MS.MacawSimulatorState sym` back to a generic  `p` type parameter.  The fixed personality was left in the few locations where it needed to be explicit.

PR #264, opened 2022-02-21
Only fix personality to (MS.MacawSimulatorState sym) where required. (take two)

RyanGlScott:
This is the same patch as #203, but rebased and with an additional commit to parameterize `Lookup{Function,Syscall}Handle` by the `personality` type variable, which proves handy in my particular use case. I've also added another commit to note these changes in the `macaw-symbolic` changelog.


travitch (approved):
I approve in general. We'll want to give the SAW folks a heads up since I believe they have some customizations for x86 that will be affected

RyanGlScott:
> We'll want to give the SAW folks a heads up since I believe they have some customizations for x86 that will be affected

Yes, good point. I'm happy to write the corresponding SAW PR.

PR #267, opened 2022-03-10
[symbolic] Optimize how initial binary memories are populated

travitch:
Before, the API provided by macaw-symbolic asserted the initial value of each byte of memory individually. This was fairly expensive for large binaries, as each such assertion flushed the solver pipe.

This change generates a large conjunction of assertions and sends them all at once. In unscientific testing, this saved half an hour on a large binary.

API Changes:

- Note that it introduces a minor API change. The optimization required that the `sym` parameter be concretely an `ExprBuilder`.


travitch:
@robdockins I tagged you on this one just to make sure you don't have additional suggestions around this change in what4 usage

robdockins (approved):

robdockins:
Nothing to add, except that we should maybe revisit the absorption rewrite generally.

PR #268, opened 2022-03-17
Add ability to load multiple memories into a flat address space

bboston7:
This change adds a function `newMergedGlobalMemoryWith`, which acts like
`newGlobalMemoryWith` but takes a list of macaw memories and merges them
into a flat address space.  This aids in reasoning dynamically linked
programs.

### DIFF
--- a/symbolic/src/Data/Macaw/Symbolic/Memory.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/Memory.hs
@@ -124,6 +124,7 @@ import qualified Data.BitVector.Sized as BV
 import qualified Data.ByteString as BS
 import qualified Data.Foldable as F
 import qualified Data.IntervalMap.Strict as IM
+import qualified Data.List.NonEmpty as NEL
 
 import qualified Data.Parameterized.NatRepr as PN
 import qualified Data.Parameterized.Context as Ctx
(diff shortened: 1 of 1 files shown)


symbolic/src/Data/Macaw/Symbolic/Memory.hs
```diff
@@ -242,25 +243,68 @@ newGlobalMemoryWith
...
+ -- ^ Hooks customizing the memory setup
+ -> proxy arch
+ -- ^ A proxy to fix the architecture
+ -> bak
+ -- ^ The symbolic backend used to construct terms
+ -> CLD.EndianForm
+ -- ^ The endianness of values in memory
+ -> MemoryModelContents
+ -- ^ A configuration option controlling how mutable memory should be represented (concrete or symbolic)
+ -> [MC.Memory (MC.ArchAddrWidth arch)]
```
travitch, line 271: I suggest a `Data.List.NonEmpty` here

symbolic/src/Data/Macaw/Symbolic/Memory.hs
```diff
@@ -242,25 +243,68 @@ newGlobalMemoryWith
...
+newMergedGlobalMemoryWith hooks proxy bak endian mmc mems = do
   let sym = CB.backendGetSym bak
-  let ?ptrWidth = MC.memWidth mem
+
+  let ?ptrWidth =
+        case map MC.memWidth mems of
+          [] -> error "Cannot build merged global memory from empty list"
+          x : xs ->
+            -- Check that all memories have the same pointer width
+            if all (testNatEq x) xs
```
travitch, line 282: Don't we already have a proof that the widths are all `ArchAddrWidth arch` (by virtue of the type of the list)? That would make the test unnecessary

symbolic/src/Data/Macaw/Symbolic/Memory.hs
```diff
@@ -242,25 +243,68 @@ newGlobalMemoryWith
  -> MC.Memory (MC.ArchAddrWidth arch)
  -- ^ The macaw memory
  -> m (CL.MemImpl sym, MemPtrTable sym (MC.ArchAddrWidth arch))
-newGlobalMemoryWith hooks proxy bak endian mmc mem = do
+newGlobalMemoryWith hooks proxy bak endian mmc mem =
+  newMergedGlobalMemoryWith hooks proxy bak endian mmc [mem]
+
+-- | A version of 'newGlobalMemoryWith' that takes a list of memories and
+-- merges them into a flat addresses space.
```
travitch, line 250: Can you document the required precondition that the address spaces must not overlap?

Also, it is worth a note that we will need to add configuration options to `GlobalMemoryHooks` to support other strategies for merging memories.

symbolic/src/Data/Macaw/Symbolic/Memory.hs
```diff
@@ -321,23 +365,24 @@ populateMemory :: ( CB.IsSymBackend sym bak
                -- ^ The symbolic backend
                -> MemoryModelContents
                -- ^ A flag to indicate how to populate the memory model based on the memory image
-               -> MC.Memory (MC.ArchAddrWidth arch)
+               -> [MC.Memory (MC.ArchAddrWidth arch)]
```
travitch, line 368: I suggest making this a `Foldable` container (so that we can pass in the `NonEmpty` list from the new entry point easily)

PR #269, opened 2022-03-17
Add RV32GC support to macaw-riscv

bboston7:
This change adds support for RV32GC RISC-V binaries.  Specifically, it:

* Updates the return matcher to recognize returns in 32-bit binaries
* Updates detection of unsupported binaries to allow RV32GC binaries
* Adds RV32GC versions of the RV64GC tests


macaw-riscv/src/Data/Macaw/RISCV/Identify.hs
```diff
@@ -62,7 +62,12 @@ matchReturn :: RISCVConstraints rv
             -> Maybe (Some (MA.AbsValue w))
 matchReturn rvRepr absProcState ip = G.withRV rvRepr $ do
   MC.BVAnd _ addr (MC.BVValue _ mask) <- MC.valueAsApp ip
-  guard (mask == 0xfffffffffffffffe)
+  let maskVal = case G.rvBaseArch rvRepr of
+                  G.RV64Repr -> 0xfffffffffffffffe
+                  G.RV32Repr -> 0xfffffffe
+                  _ -> error ( "Unsupported base architecture: "
+                            ++ (show (G.rvBaseArch rvRepr)))
```
travitch, line 69: What do you think about converting the mask into a `BV w` from bv-sized and then making `maskVal = sext w (mkBV (knownNat @8) 0xfe)`

That would be resilient to any architecture width.

macaw-riscv/src/Data/Macaw/RISCV/Identify.hs
```diff
@@ -62,7 +62,12 @@ matchReturn :: RISCVConstraints rv
             -> Maybe (Some (MA.AbsValue w))
```
travitch, line 62: Do we get to fix the FIXME comment on this function with this change?

travitch (approved):

macaw-riscv/src/Data/Macaw/RISCV/Identify.hs
bboston7: Done!

macaw-riscv/src/Data/Macaw/RISCV/Identify.hs
bboston7: Yep!  Good catch.  I removed the FIXME.

PR #284, opened 2022-05-02
Reverse syscall override return register ordering

bboston7:
When a user overrides a system call on an architecture that supports returning two values from a system call and they provide a context containing the result of the system call in the form

```
empty :> v0 :> v1
```

macaw will perform the register assignment

```
r0 := v1
r1 := v0
```

This change reverses this behavior so that the assignment becomes

```
r0 := v0
r1 := v1
```

This brings the expected ordering of the result context in agreement
with the left-to-right ordering of the argument context:

```
empty :> arg1 :> arg2 :> ...
```

### DIFF
--- a/macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
+++ b/macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
@@ -644,7 +644,9 @@ a32InstructionMatcher (ARMDis.Instruction opc operands) =
                                  <*> G.getRegVal (ARMReg.ARMGlobalBV (ASL.knownGlobalRef @"_R6"))
                                  <*> G.getRegVal (ARMReg.ARMGlobalBV (ASL.knownGlobalRef @"_R7"))
             res <- G.addExpr =<< evalArchFn sc
-            -- res is a tuple of form (R1, R0)
+            -- res is a tuple of form (R1, R0).  This is reversed from the
+            -- user provided return Assignment of empty :> R0 :> R1 because
+            -- the conversion from Assignment to Tuple reversed the order.
             G.setRegVal r1 =<< G.addExpr (G.AppExpr (MC.TupleField PC.knownRepr res PL.index0))
             G.setRegVal r0 =<< G.addExpr (G.AppExpr (MC.TupleField PC.knownRepr res PL.index1))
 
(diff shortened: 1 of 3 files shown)


macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
```diff
@@ -698,8 +699,9 @@ t32InstructionMatcher (ThumbDis.Instruction opc operands) =
                                    <*> G.getRegVal (ARMReg.ARMGlobalBV (ASL.knownGlobalRef @"_R6"))
                                    <*> G.getRegVal (ARMReg.ARMGlobalBV (ASL.knownGlobalRef @"_R7"))
             res <- G.addExpr =<< evalArchFn sc
-            G.setRegVal r0 =<< G.addExpr (G.AppExpr (MC.TupleField PC.knownRepr res PL.index0))
-            G.setRegVal r1 =<< G.addExpr (G.AppExpr (MC.TupleField PC.knownRepr res PL.index1))
+            -- res is a tuple of form (R1, R0)
+            G.setRegVal r1 =<< G.addExpr (G.AppExpr (MC.TupleField PC.knownRepr res PL.index0))
+            G.setRegVal r0 =<< G.addExpr (G.AppExpr (MC.TupleField PC.knownRepr res PL.index1))
```
RyanGlScott, line 704: Ah, now I can see why this was so easy to overlook—it's because we're using `index0` and `index1` in what seems like an intuitive order, but the result doesn't match that intuition. Perhaps we should explicitly match on `res` rather than using indices to make this more obvious, e.g.,

```hs
let r1 :> r0 :> Nil = res
G.setRegVal r1
G.setRegVal r2
```

It might also be worth writing a Note to explain why we end up with seemingly the opposite order you'd expect for returned values, explaining the impedance mismatch between `Data.Parameterized.List` and `Assignment`.

travitch: Agreed

travitch:
I just want to note that "result value" and "error value" is not necessarily an accurate characterization. There are simply some syscalls that return two values (though very few).

travitch (approved):
Approved, aside from the documentation request

macaw-aarch32/src/Data/Macaw/ARM/Arch.hs
bboston7: The pattern match doesn't work because `res` is an `Expr ids ('TupleType '[BVType 64, BVType 64])` and not just a `Data.Parameterized.List`.  Getting to the individual list would require evaluating the `Expr`, converting the `Value` into a `PL.List` (this looks potentially gross), matching on the list, then building up new `Expr`s for each field.  I think this could be more confusing than the current approach as the match on the list will be hidden among a bunch of evaluation/extraction/wrapping logic.

I will definitely add a comment about the impedance mismatch though.

bboston7:
> I just want to note that "result value" and "error value" is not necessarily an accurate characterization. There are simply some syscalls that return two values (though very few).

Thanks.  I updated the description to reflect this.

PR #291, opened 2022-05-23
Submodule updates

travitch:
The primary changes are version bounds to let everything build against the latest what4 release (1.3)


RyanGlScott:
I suspect that you're going to run into trouble when bumping the `what4` submodule because its test suite requires `tasty-hedgehog >= 1.2`, but `bv-sized`'s test suite currently requires `tasty-hedgehog < 1.2`. See GaloisInc/bv-sized#23. I suspect that the error [here](https://github.com/GaloisInc/macaw/runs/6564187906?check_suite_focus=true#step:22:1) is a result of this:

```
Run cabal test pkg:macaw-refinement
cabal: Cannot test the package macaw-refinement-0.0.1 because none of the
components are available to build: the test suite 'test-refinements' is not
available because the solver did not find a plan that included the test
suites. Force the solver to enable this for all packages by adding the line
'tests: True' to the 'cabal.project.local' file.
```

PR #293, opened 2022-06-07
Add tests for 32 bit PowerPC

### LINKED ISSUE #227: Mis-classification of returns on PowerPC 32
In an example for another project[1], macaw fails to classify the return because it seems to lose track of the return address on the stack after a sequence of local calls.

[1] https://github.com/GaloisInc/pate/blob/master/tests/ppc32/test-call-twice.original.exe

travitch:
Add a test suite for PowerPC 32. There was already one for PowerPC 64. This change generalizes the test harness to support both in one codebase. It also adds support for a `--save-macaw` flag to the test harness to save discovered macaw IR to a file in /tmp to ease debugging.

This addresses a few classification errors in PowerPC 32 that seemingly don't affect PowerPC 64:
- A mishandling of abstraction of the link register after a system call

Note that this change includes a Dockerfile for building the tests; it is trickier than average because PowerPC 32 cross compilers are not as commonly available as other architectures.

Fixes #227


macaw-ppc/tests/PPC64Tests.hs
```diff
@@ -1,42 +1,55 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
 module PPC64Tests (
```
RyanGlScott, line 10: Perhaps this should be renamed to just `PPCTests`, since this is no longer specific to 64-bit PPC.

macaw-ppc/tests/Main.hs
```diff
@@ -1,16 +1,27 @@
...
+ingredients = T.includingOptions [ TTO.Option (Proxy @PPC64.SaveMacaw)
+                                 ] : T.defaultIngredients
+
+
 main :: IO ()
 main = do
-  testFiles <- namesMatching "tests/ppc/*.s.expected"
+  testFiles64 <- namesMatching "tests/ppc/*.s.expected"
+  testFiles32 <- namesMatching "tests/ppc32/*.s.expected"
   bins <- namesMatching "tests/ppc/bin/*"
```
RyanGlScott, line 22: Is everything under `tests/ppc` related to 64-bit PPC? If so, should it be renamed to `tests/ppc64` for symmetry with the new `tests/ppc32` subdirectory?

RyanGlScott, line 22: Is everything under `tests/ppc` related to 64-bit PPC? If so, should it be renamed to `tests/ppc64` for symmetry with the new `tests/ppc32` subdirectory?

base/src/Data/Macaw/AbsDomain/AbsState.hs
```diff
@@ -611,6 +638,14 @@ asBoolConst (BoolConst b) = Just b
 asBoolConst TopV = Nothing
 
 -- | Add an integer to the abstract value.
+--
+-- This is not exported and is only used to increment or decrement the stack
+-- pointer by a known (by the architecture backend) signed integer after a call
+-- returns. The integer is *not* a bitvector.
```
RyanGlScott, line 644: I'm a bit confused by the statement "The integer is *not* a bitvector", as the integer is passed directly to `addSignedOffsetBV` in some places, whose Haddocks emphatically state that the integer _is_ a bitvector. Which is it?

base/src/Data/Macaw/AbsDomain/AbsState.hs
```diff
@@ -93,8 +92,36 @@ import           Data.Macaw.Types
...
 
-addOff :: NatRepr w -> Integer -> Integer -> Integer
-addOff w o v = toUnsigned w (o + v)
+-- | A helper for computing new signed offsets for values of type 'StackOffsetAbsVal'
+--
+-- This is a bit fiddly because 'StackOffsetAbsVal' has an 'Int64' as its second
+-- value, so we need to be very careful when converting from bitvectors (encoded
+-- as 'Integer') to 'Int64'.
+addSignedOffset
+  :: NatRepr w
```
RyanGlScott, line 101: What is the purpose of this `NatRepr w` argument? It's not used anywhere in the implementation. Is the idea that the `Integer` should be a bitvector of length `w`? If so, it might be worth stating this in the Haddocks.

RyanGlScott, line 101: What is the purpose of this `NatRepr w` argument? It's not used anywhere in the implementation. Is the idea that the `Integer` should be a bitvector of length `w`? If so, it might be worth stating this in the Haddocks.

base/src/Data/Macaw/AbsDomain/AbsState.hs
```diff
@@ -93,8 +92,36 @@ import           Data.Macaw.Types
...
+--
+-- This handles the interpretation of the bitvector as an 'Int64'. It uses the
+-- signed conversion function from the 'Data.Parameterized.NatRepr' module. This
+-- takes the low @w@ bits of the Integer and represents it in twos-complement so
+-- that the conversion to an 'Int64' properly sign extends.
+addSignedOffsetBV
+  :: (1 <= w)
+  => NatRepr w
+  -> Int64 -- ^ The current offset of a 'StackOffsetAbsVal'
+  -> Integer -- ^ A *bitvector* of width @w@ represented as an 'Integer' to add to the current offset; it will be interpreted as signed
```
RyanGlScott, line 121: Would it not be more direct just to use a `BV w` here instead of an `Integer`? Half of `addSignedOffsetBV`'s call sites have access to a `BV` anyway, so it doesn't look like you're saving yourself that many keystrokes by using an `Integer` here.

RyanGlScott, line 121: Would it not be more direct just to use a `BV w` here instead of an `Integer`? Half of `addSignedOffsetBV`'s call sites have access to a `BV` anyway, so it doesn't look like you're saving yourself that many keystrokes by using an `Integer` here.

travitch: I don't think anything in this module as access to a `BV`, does it? I aspire to translate this module to use `BV` for safety and avoid this class of problems more systematically, but this was the module we skipped when doing the port of macaw to bv-sized because the changes would be so pervasive and scary without some unit tests.

base/src/Data/Macaw/AbsDomain/AbsState.hs
travitch: Hmm this comment is referring specifically to the `Integer` argument to `bvinc`, which is indeed not a bitvector. It is only passed to `addSignedOffset` (the non-BV) version.

This is one of the few `Integer`s in the module that is not a bitvector, so I wanted to call it out (since I was initially worried about how to handle it while making this change, but then realized that it was the safe and easy case).

I can clarify that it is being called out because this is an exception to the normal order of business in the module

RyanGlScott (approved):

PR #304, opened 2022-07-19
`macaw-symbolic`: Generalize some uses of `CrucibleState` to `SimState`

RyanGlScott:
This patch was motivated by the need to call `doGetGlobal` from a Crucible override, where the `SimState` is instantiated with `OverrideLang` rather than `CrucibleLang`, the latter of which is used in the `CrucibleState` type synonym. While I was in town, I generalized the types of other operations in `Data.Macaw.Symbolic.MemOps` where it was reasonable.


travitch (approved):
Nice, love to see it

Issue #300, opened 2022-07-11
Macaw symbolic seems to produce doubled memory safety assertions


robdockins:
When doing verification using macaw symbolic via SAW, I'm getting what appear to be redundant proof obligations on memory reads and writes.  The resulting proof obligations always seem to be paired up, like the following:

```
[16:52:52.607] Goal aesni_gcm_decrypt (goal number 0): safety assertion
at 0x76031e in aesni_gcm_decrypt
[doReadMem] Reading through an invalid pointer: (5, 0x50:[64])


[16:52:52.651] WARNING: admitting goal aesni_gcm_decrypt
[16:52:52.658] Goal aesni_gcm_decrypt (goal number 1): safety assertion
at 0x76031e in aesni_gcm_decrypt
Error during memory load
```

This appears to be caused by these (And similar lines elsewhere)

https://github.com/GaloisInc/macaw/blob/2791b1050f348922a27192d8a73ee538fed57629/symbolic/src/Data/Macaw/Symbolic/MemOps.hs#L1019-L1022

The `assertSafe` will generate one obligation, and the `fail` also appears to bubble out and generate a separate proof obligation.

Are these actually asserting separate things that are necessary for soundness? If so, can we combine them somehow into a single obligation? If not, what's the best way to remove one of them?

robdockins:
On further investigation, I think the additional assertion is actually coming from here:

https://github.com/GaloisInc/macaw/blob/2791b1050f348922a27192d8a73ee538fed57629/symbolic/src/Data/Macaw/Symbolic/MemOps.hs#L1008-L1012

@travitch, as far as I can tell, this check is just redundant (as are related ones in doWriteMem, etc.) is there some aspect of this that I am missing?

travitch:
@robdockins Is it redundant because `loadRaw` and similar functions ultimately do their own checks?

I'm not that familiar with the reasoning for why these functions are structured the way they are. I suspect you are right. If I had to guess, maybe these were copied and pasted from crucible-llvm before a refactoring that cleaned things up?

If `loadRaw` and friends already do this check, I have no objections to removing it. My only concern would be if some downstream tools are counting the number of assertions and relying on that somehow. I have one that I'm happy to fix. Perhaps @andreistefanescu or @chameco know if any of the x86 SAW proofs might be disrupted?

robdockins:
[...]

PR #310, opened 2022-08-03
Remove redundant "valid pointer" checks from memory loads and stores.

robdockins:
The internal memory model operations already assert the same checks.

CF #300


robdockins:
SAW CI is passing with this change, so I think we are probably good on that front.

https://github.com/GaloisInc/saw-script/pull/1722

travitch (approved):

robdockins:
Should I be concerned about the GHC 9.2.2 build fail? I don't see how it could be related to this.

travitch:
No, it is fine. Looks like the runner crashed again. Feel free to merge when you feel like it

Issue #47, opened 2019-06-02
Support R_X86_64_COPY


joehendrix:
Elf relocations of type `R_X86_64_COPY` are used in dynamic binaries that reference variables defined in shared libraries.  We should add support for these, but it may be a bit tricky.

The variables are stored in the binary's `.bss` section, and so the relocation target is within the `.bss` section for the binary.  These relocations can be used to tell us that address is used to store a given global variable, and this information may be of use in LLVM generation or other binaries transformations.

A couple references that describe this are <https://www.cs.stevens.edu/~jschauma/631/elf.html> and <https://stac47.github.io/c/relocation/elf/tutorial/2018/03/01/understanding-relocation-elf.html>.

In addition to adding support, test cases should be written to validate the understanding.

simonjwinwood:
This is the updated link: https://stevens.netmeister.org/631/elf.html

One problem with the ```R_X86_64_COPY``` relocation is that the size of the relocation depends on the size of the symbol.  This relocation is used to copy the value of an initialised variable into the location in the binary where it lives --- the dynamic library will have a copy of the variable which is the source of the copy.

PR #311, opened 2022-09-13
`macaw-base`: Support loading `R_ARM_COPY` ELF relocations

RyanGlScott:
We already have support for `R_X86_64_COPY` relocations, so adding support for `R_ARM_COPY` on the AArch32 side is straightforward.

This is related to #47, although this is not a full fix for the issue.

### DIFF
--- a/base/src/Data/Macaw/Memory/ElfLoader.hs
+++ b/base/src/Data/Macaw/Memory/ElfLoader.hs
@@ -579,7 +579,7 @@ relaTargetX86_64 _ symtab rel addend _isRel =
     Elf.R_X86_64_COPY -> do
       sym <- resolveRelocationSym symtab (Elf.relSym rel)
       when (addend /= 0) $ do
-        throwError$ RelocationUnsupportedType (show (Elf.relType rel))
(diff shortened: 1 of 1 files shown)


base/src/Data/Macaw/Memory/ElfLoader.hs
```diff
@@ -690,6 +690,18 @@ relaTargetARM32 end msegIndex symtab rel addend relFlag =
...
+      when (addend /= 0) $ do
+        throwError $ RelocationUnsupportedType (show (Elf.relType rel))
+      pure $ Relocation { relocationSym        = sym
+                        , relocationOffset     = 0
+                        , relocationIsRel      = False
+                        , relocationSize       = 4
+                        , relocationIsSigned   = False
+                        , relocationEndianness = end
+                        , relocationJumpSlot   = False
+                        }
```
RyanGlScott, line 704: This code is almost entirely copy-pasted from the `R_X86_64_COPY` case above, except that I changed the `relocationSize` from `8` to `4`. I believe the other fields don't change for AArch32.

bboston7 (approved):

Issue #302, opened 2022-07-13
ELF loader crashes on `arm-linux-gnueabi-gcc` binary (`Found section header that overlaps with program header.`)


RyanGlScott:
Take the following program (adapted from https://github.com/GaloisInc/macaw/issues/301#issuecomment-1183555233):

```c
// test.c
#include <stdint.h>

uint64_t x = 0;

int __attribute__((noinline)) test_strd() {
  x = 42;
  return x == 42;
}

int main() {
  test_strd();
  return 0;
}
```

And compile it with `arm-linux-gnueabi-gcc` like so:

```
$ arm-linux-gnueabi-gcc -nostartfiles -O2 -static test.c -o test.exe
```

And run it through this test harness:

<details>

```hs
-- Bug.hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE FlexibleContexts #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE ImplicitParams #-}
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TypeApplications #-}
{-# LANGUAGE TypeOperators #-}
module Main (main) where

import           Control.Lens ( (^.) )
import qualified Data.ByteString as BS
import qualified Data.ByteString.Char8 as BS8
import qualified Data.ElfEdit as Elf
import qualified Data.Foldable as F
import qualified Data.Map as Map
import           Data.Maybe ( mapMaybe )
import qualified Data.Parameterized.Classes as PC
import qualified Data.Parameterized.Nonce as PN
import           Data.Parameterized.Some ( Some(..) )
import           Data.Proxy ( Proxy(..) )

import qualified Language.ASL.Globals as ASL
import qualified Data.Macaw.Architecture.Info as MAI

import           Data.Macaw.AArch32.Symbolic ()
import qualified Data.Macaw.ARM as MA
import qualified Data.Macaw.ARM.ARMReg as MAR
import qualified Data.Macaw.Discovery as M
import qualified Data.Macaw.Symbolic as MS
import qualified Data.Macaw.Symbolic.Testing as MST
import qualified What4.Config as WC
import qualified What4.Expr as WE
import qualified What4.Expr.Builder as WEB
import qualified What4.Interface as WI
import qualified What4.ProblemFeatures as WPF
import qualified What4.Solver as WS

import qualified Lang.Crucible.Backend as CB
[...]

PR #317, opened 2023-02-15
workaround for IntervalMap empty intersections

danmatichuk:
The function IntervalMap.intersecting is incorrectly returning nonempty results when given an empty interval. This works around the issue by explicitly adding an isEmpty check.

This resolves an issue where ElfLoader would error, saying "Found section header that overlaps with program header.", if presented with a segment that has zero file size.

### DIFF
--- a/base/src/Data/Macaw/Memory/ElfLoader.hs
+++ b/base/src/Data/Macaw/Memory/ElfLoader.hs
@@ -1181,6 +1181,8 @@ insertElfSegment regIdx addrOff shdrMap contents relocMap phdr = do
     mlsMemory %= memBindSegmentIndex segIdx seg
     -- Iterative through sections
     let interval = IntervalCO phdrOffset phdrEnd
+    -- Workaround for https://github.com/bokesan/IntervalMap/issues/9, where 'IMap.intersecting' can return
+    -- nonempty results when given an empty interval.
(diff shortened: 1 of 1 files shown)


RyanGlScott:
> The function IntervalMap.intersecting is incorrectly returning nonempty results when given an empty interval.

Huh. Can you provide an example of some `IntervalMap` code that returns incorrect results? It sounds like we ought to file a bug report against `IntervalMap`, assuming that someone else hasn't already done so before.

danmatichuk:
I've posted an issue to the github project for `IntervalMap` here https://github.com/bokesan/IntervalMap/issues/9, which contains an example.

RyanGlScott:
Thanks! I think it would be worth leaving a comment next to the code that explains why we bother to include this unusual special case.

Also, I did some limited testing, and this issues appears to fix #302! Can you add that as a test case? Alternatively, perhaps you have a test case of your own?

RyanGlScott:
Now that `IntervalMap-0.6.2.0` is on Hackage with a fix for bokesan/IntervalMap#9, perhaps we should just bump the lower version bounds on `IntervalMap` to `>= 1.6.2`. That way, we can drop the workaround entirely.

danmatichuk:
Agreed. I'll roll back this change in favor of jump bumping the revision and then add your test to the repo.

PR #319, opened 2023-02-16
support hashable-1.4

Ptival:
Either we want to put an explicit upper bound on hashable (which is currently not explicitly listed, but is depended upon transitively via parameterized-utils re-exporting `Hashable` in `Data.Parameterized.Classes`), or we can sprinkle in a couple of `Eq` instances like so.

Not sure who to review this, but since @RyanGlScott you've been doing such changes elsewhere, maybe you have an opinion on this!

### DIFF
--- a/base/ChangeLog.md
+++ b/base/ChangeLog.md
@@ -21,3 +21,7 @@
 - The `DynamicSymbolTable` constructor of `Data.Macaw.Memory.ElfLoader`'s
   `SymbolTable` data type now has an additional `VersionDefMap` field, which is
   needed for finding versioning information in some cases.
+
+- The `Hashable` and `HashableF` instances for `App f` now require
+  `TestEquality f` constraints. (This is needed to support `hashable-1.4.*`,
+  which adds `Eq` as a superclass to `Hashable`.)
--- a/base/src/Data/Macaw/CFG/App.hs
+++ b/base/src/Data/Macaw/CFG/App.hs
@@ -385,7 +385,7 @@ instance TestEquality f => TestEquality (App f) where
                    ]
                   )
 
-instance HashableF f => Hashable (App f tp) where
+instance (HashableF f, TestEquality f) => Hashable (App f tp) where
   hashWithSalt = $(structuralHashWithSalt [t|App|]
                      [ (DataArg 0 `TypeApp` AnyType, [|hashWithSaltF|])
                      , (ConType [t|TypeRepr|] `TypeApp` AnyType, [|\s _c -> s|])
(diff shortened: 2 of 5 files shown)


RyanGlScott:
I agree that we eventually want to support `hashable-1.4.*`. That being said, not all of the downstream libraries that the `macaw` repo depends on currently support `hashable-1.4.*`, as can be seen when you run this command in the `macaw` repo:

```
$ cabal build macaw-base --constraint="hashable==1.4.*"
Resolving dependencies...
Error: cabal: Could not resolve dependencies:
[__0] trying: asl-parser-0.1.0.0 (user goal)
[__1] trying: base-4.16.3.0/installed-4.16.3.0 (dependency of asl-parser)
[__2] trying: asl-translator-0.1.0.0 (user goal)
[__3] trying: libBF-0.6.5.1 (dependency of asl-translator)
[__4] trying: hashable-1.4.2.0 (dependency of libBF)
[__5] trying: parameterized-utils-2.1.6.0 (dependency of asl-translator)
[__6] trying: hashtables-1.3.1 (dependency of asl-translator)
[__7] next goal: semmc (user goal)
[__7] rejecting: semmc-0.1.0.0 (conflict: hashtables==1.3.1, semmc =>
hashtables>=1.2 && <1.3)
[__7] fail (backjumping, conflict set: hashtables, semmc)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: hashable, hashtables,
parameterized-utils, asl-translator, semmc, base, libBF, asl-parser
Try running with --minimize-conflict-set to improve the error message.
```

We should [patch `semmc` first to support `hashtables-1.3.*`](https://github.com/GaloisInc/semmc/blob/f7d6d856ad930aaad534a7df0beb39a13be2423f/semmc/semmc.cabal#L86), which is the first release in the series that supports `hashable-1.4.*`. Then, we can update the `semmc` module here so that we can be sure that everything builds against `hashable-1.4.*`. (I'm not sure if there are any other libraries that would need similar treatment.)

RyanGlScott:
I've opened a PR to make the relevant changes on the `semmc` side at GaloisInc/semmc#77.

RyanGlScott:
I've pushed a change which bumps the `semmc` submodule to bring in the changes from GaloisInc/semmc#77. Once the CI passes, I'll land this. Thanks for the patch, @Ptival!

RyanGlScott (approved):

Ptival:
Alright, let's land this. If this breaks anyone's workflow, reach out to me! :-)

Issue #282, opened 2022-04-25
`macaw-symbolic`: Lazily initialize global memory


RyanGlScott:
Currently, `macaw-symbolic`'s memory model creates an SMT array to back global memory and then immediately adds assertions relating to every byte in the memory. Doing all this work upfront is expensive, especially for large binaries. What's more, most binaries don't actually need to make use of the entirety of global memory, so many of these assertions will go unused.

Rather than asserting everything upfront, a more scalable approach would be to split the global memory up into smaller regions and only add assertions for each region on demand. The `MemPtrTable` data type already has much of the information it needs to do this, since it contains an `IntervalMap` that is divided up into `MemChunk`s. We'd want to be even more fine-grained than that, however, since binaries will often have large `MemChunk`s that can incur bad overhead when symbolically initialized. In order to track this state, we may need to track additional state in `MacawSimulatorState`.

In addition to lazily populating the symbolic bytes, another optimization opportunity is to avoid using the SMT array entirely when performing concrete reads from read-only memory. Such memory can never be changed during simulation, so there's no point using a (comparatively expensive) SMT array to track it.

RyanGlScott:
We should be careful and not require that all `macaw-symbolic` users make use of this particular memory layout, if for no other reason than SAW has a very bespoke memory layout that is more integrated with LLVM. I will need to think about what the right abstraction boundaries are here so that we can offer enough hooks into `macaw-symbolic`'s memory model to make use of a lazy memory model while still permitting SAW to do its own thing.

PR #327, opened 2023-03-09
`macaw-symbolic`: Add alternative, lazy memory model

### LINKED ISSUE #323: `macaw-symbolic`: Concretize pointers in read/writes when possible
One particularly noticeable performance bottleneck in `macaw-symbolic` is reading from or writing to symbolic pointers in a `MacawReadMem` or `MacawWriteMem` statement, as this contributes to large mux trees in the queries that are sent to SMT solvers. While it is not possible to avoid symbolic pointers in general, it is often the case that a pointer being read from or written to is:

1. Symbolic, but easily resolved to be concrete using `what4`'s [`resolveSymBV`](https://github.com/GaloisInc/what4/blob/ea717ac94a186b5ee18f138b71d8b4b4b2f00955/what4/src/What4/Utils/ResolveBounds/BV.hs#L86-L107) function.
2. Symbolic, but with a small range of possible values allowed. In this case, `what4`'s [`unsafeSetAbstractValue`](https://github.com/GaloisInc/what4/blob/98ae759128a5c3de947b59730806b110ec8c005d/what4/src/What4/Interface.hs#L391-L407) function to update the bounds of the pointer offset, which may unlock further `what4` simplifications.

I have implemented this approach in a downstream project to great success, in some cases reducing the overall time it takes to run certain test cases by half. The flip side to this approach is that:

* `resolveSymBV` requires the use of an `OnlineSolver` constraint, which is an assumption that `macaw-symbolic` does not currently make.
* While this approach generally makes the `macaw-symbolic` memory model faster, it is possible that one could construct a program where repeated uses of `resolveSymBV` turn it into a performance bottleneck. (I haven't seen any programs in the wild that do this, but I wouldn't discount the possibility.)

As a result, I think we should design this so that this option is guarded behind an optional configuration of some kind.

RyanGlScott:
The current `macaw-symbolic` memory model has issues when scaling up to binaries that have several megabytes or more in size. This patch introduces a new memory model (in `Data.Macaw.Symbolic.Memory.Lazy`) that serves as a mostly drop-in replacement for the existing memory model (which I now refer to as the "default" memoy model). The lazy memory model scales better by incrementally populating the SMT array backing global memory over the course of a run of the simulator. For the full details, see `Note [Lazy memory model]`. I performed some refactoring to share common bits between the default and lazy memory models. This fixes https://github.com/GaloisInc/macaw/issues/282.

Another component this is needed to make the lazy memory model perform better is concretizing pointers during a read or write. The details of how concretization work can be found in the new `Data.Macaw.Symbolic.Concretize` module. Whether to concretize pointers or not is controlled behind an option in the new `MemModelConfig` data type. This fixes #323.

Finally, this folds a variety of assorted `macawExtensions` arguments into the `MemModelConfig` data type. The advantage to doing this is that is will make it easier to extend the memory model configuration options in the future without needlessly foisting breaking changes on all `macaw-symbolic` users. Unfortunately, it does require a breaking change to get to this point, but the migration path is straightforward for existing code. I have included this migration story in the `macaw-symbolic` changelog.


RyanGlScott:
> The `MemModelConfig` is a nice change; that could perhaps have been a precursor PR to result in a smaller change, but it is a separate commit, allowing bisection in the event of issues.

Yes, I will admit that the order of commits in this PR is somewhat strange, since I fold everything into `MemModelConfig` in the last commit in the series, rather than doing that first. (I only fully committed to this design after making the first two commits, which made it more obvious what the design should be.) If you'd like, I can split out the `MemModelConfig` changes and make that the first commit in the series. I could also make it a separate PR, although it wouldn't make much of a difference vis-à-vis bisectability.

> The testing appears to be focused on the new lazy memory model. Is the full memory model being sufficiently tested with this, or should there be some additional testing introduced to ensure coverage of both?

A good question. The only test coverage that the default memory model is receiving after this patch is via `macaw-refinement`, which is probably not ideal. I suppose we should tweak `Data.Macaw.Symbolic.Testing` to run each test case with both the default and lazy memory models... I will have to think about the best way to do that.

> The changelog should mention the lazy memory model.

Will do!

symbolic/src/Data/Macaw/Symbolic/Memory/Lazy.hs
```diff
@@ -0,0 +1,826 @@
...
+-- with qualified imports.
+module Data.Macaw.Symbolic.Memory.Lazy (
+  -- * Memory Management
+  memModelConfig,
+  MemPtrTable,
+  MSMC.toCrucibleEndian,
+  MSMC.fromCrucibleEndian,
+  MSMC.GlobalMemoryHooks(..),
+  MSMC.defaultGlobalMemoryHooks,
+  newMergedGlobalMemoryWith,
```
RyanGlScott, line 32: Note to self: we should also offer versions of `newGlobalMemory` and `newGlobalMemoryWith` for symmetry with `Data.Macaw.Symbolic.Memory`'s API.

RyanGlScott:
Looking at the code more closely, I think it would be straightforward to add an additional option to the `simulateAndVerify` command (the main entrypoint to testing `macaw-symbolic`) that controls which memory model to use. That being said, I'm unsure how far we should go with this. Do we want to run every test case with both memory models, effectively doubling the time it takes to run the `macaw-symbolic`–related test suites, or do we only want to run a subset of the test cases with a particular memory model?

kquick:
It looks like the tests take a trivial amount of time, so I like the idea of parameterizing `simulateAndVerify` and allowing both models to be fully tested.

kquick:
> I could also make it a separate PR, although it wouldn't make much of a difference vis-à-vis bisectability.

I don't think it's worth it at this point.  I had mistaken the order of the commits, but I don't think there's a significant concern about future bisection to that fine of a degree.

PR #333, opened 2023-07-28
applying reasonable hlint suggestions

Ptival:
These are the least controversial hlint suggestions.

In particular, cleaning up unused pragmas, and redundant syntax that does not help with understanding.

Feel free to disagree with any subset of these!

### DIFF
--- a/base/src/Data/Macaw/Discovery/State.hs
+++ b/base/src/Data/Macaw/Discovery/State.hs
@@ -7,7 +7,6 @@ representing this information.
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE UndecidableInstances #-}
 module Data.Macaw.Discovery.State
   ( -- * DiscoveryState
@@ -240,8 +239,8 @@ emptyDiscoveryState mem addrSymMap info =
   , _globalDataMap       = Map.empty
   , _funInfo             = Map.empty
   , _unexploredFunctions = Map.empty
-  , _trustedFunctionEntryPoints = fmap (\_ -> Info.MayReturnFun) addrSymMap
-  , _exploreFnPred       = \_ -> True
+  , _trustedFunctionEntryPoints = Info.MayReturnFun <$ addrSymMap
+  , _exploreFnPred       = const True
   }
 
 -- | Map each jump table start to the address just after the end.
--- a/base/src/Data/Macaw/Memory.hs
+++ b/base/src/Data/Macaw/Memory.hs
@@ -13,7 +13,6 @@ Declares 'Memory', a type for representing segmented memory with permissions.
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableSuperClasses #-}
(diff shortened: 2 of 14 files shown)


base/src/Data/Macaw/Discovery/State.hs
```diff
@@ -240,8 +239,8 @@ emptyDiscoveryState mem addrSymMap info =
   , _globalDataMap       = Map.empty
   , _funInfo             = Map.empty
   , _unexploredFunctions = Map.empty
-  , _trustedFunctionEntryPoints = fmap (\_ -> Info.MayReturnFun) addrSymMap
-  , _exploreFnPred       = \_ -> True
+  , _trustedFunctionEntryPoints = fmap (const Info.MayReturnFun) addrSymMap
```
RyanGlScott, line 242: I'm surprised that `hlint` didn't recommend using this instead:

```suggestion
  , _trustedFunctionEntryPoints = Info.MayReturnFun <$ addrSymMap
```

`fmap (const Info.MayReturnFun)` would work fine too, but the former is more efficient for some `Functor`s.

base/src/Data/Macaw/Memory.hs
```diff
@@ -421,10 +421,10 @@ memInt =
     Addr64 -> \x -> MemInt x
 
 instance Eq (MemInt w) where
-  (==) = \x y -> memIntValue x == memIntValue y
+  (==) x y = memIntValue x == memIntValue y
```
RyanGlScott, line 424: A very minor thing, but if we're going to bother eta-expanding a function that's meant to be used infix, we might as well define it infix as well:

```suggestion
  x == y = memIntValue x == memIntValue y
```

base/src/Data/Macaw/Memory.hs
```diff
@@ -446,9 +446,9 @@ instance MemWidth w => Bounded (MemInt w) where
       Addr64 -> MemInt maxBound
 
 instance MemWidth w => Num (MemInt w) where
-  (+) = \x y -> memInt $ memIntValue x + memIntValue y
-  (-) = \x y -> memInt $ memIntValue x - memIntValue y
-  (*) = \x y -> memInt $ memIntValue x * memIntValue y
+  (+) x y = memInt $ memIntValue x + memIntValue y
+  (-) x y = memInt $ memIntValue x - memIntValue y
+  (*) x y = memInt $ memIntValue x * memIntValue y
```
RyanGlScott, line 451: Similarly,

```suggestion
  x + y = memInt $ memIntValue x + memIntValue y
  x - y = memInt $ memIntValue x - memIntValue y
  x * y = memInt $ memIntValue x * memIntValue y
```

base/src/Data/Macaw/Types.hs
```diff
@@ -23,6 +21,7 @@ The type of machine words, including bit vectors and floating point
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE TypeSynonymInstances #-}
 {-# LANGUAGE UndecidableInstances #-}
+
```
RyanGlScott, line 24: Does `hlint` really enforce having a space between the language extensions and the `module` line? This strikes me as odd, since there is no space between the two in other modules (e.g., `Data.Macaw.Discovery.State`), and yet `hlint` doesn't complain about those.

I don't really have a strong opinion one way or the other, but I do find the situation strange. (Similar comments apply to other modules in this PR where a space has been inserted.)

RyanGlScott (approved):
While I am generally leery of `hlint` (it is usually too opinionated for my tastes), most of the changes here are straightforward. I've left comments on a small handful of changes that made me do a double-take.

base/src/Data/Macaw/Types.hs
Ptival: Oops, these are just compulsive by me, will revert.

PR #335, opened 2023-08-09
Pretty instance for BlockExploreReason

Ptival:


base/src/Data/Macaw/Discovery/ParsedContents.hs
```diff
@@ -65,6 +65,15 @@ data BlockExploreReason w
 
   deriving (Eq, Show)
 
+instance MemWidth w => PP.Pretty (BlockExploreReason w) where
+  pretty (NextIP b) = "next address after block " <> PP.pretty b
+  pretty FunctionEntryPoint = "function entry point"
+  pretty (SplitAt o prior) =
+    PP.vcat
+      [ "split at offset " <> PP.pretty o <> ", previously reachable from:"
```
RyanGlScott, line 73: Alternatively, if you don't want to manually insert whitespace:

```suggestion
      [ "split at offset" PP.<+> PP.pretty o <> ", previously reachable from:"
```

RyanGlScott (approved):

PR #338, opened 2023-08-10
fix CI slowness issues

Ptival:
One issue was that we were computing the restore key on the hash of a file before creating said file when restoring, but after creating the file when saving, so there was a constant mismatch.

The second issue is that we were not caching dist-newstyle, so we would always rebuild our "local" dependencies such as binary-symbols, elf-edit, etc.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,15 +44,6 @@ jobs:
         ghc-version: ${{ matrix.ghc-ver }}
         cabal-version: ${{ matrix.cabal }}
 
-    - uses: actions/cache/restore@v3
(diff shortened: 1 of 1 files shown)


RyanGlScott (approved):
I'd be worth restarting one of these jobs after it finishes to see if the caching does indeed work. If so, ship it!

Ptival:
Oh yeah I had tried it prior, e.g. here: https://github.com/GaloisInc/macaw/actions/runs/5822894123

Ptival:
10 minute run on a re-run:
https://github.com/GaloisInc/macaw/actions/runs/5824153994

PR #342, opened 2023-09-08
fix incorrect documentation

Ptival:


RyanGlScott (approved):
Wow, I have no idea how the documentation managed to get so out of sync for these functions. The new Haddocks look correct (certainly more so than the old Haddocks).

danmatichuk (approved):

PR #353, opened 2023-12-07
remove unused extension

Ptival:


RyanGlScott (approved):
I'm fine with this, although I'm not confident that we can avoid re-introducing redundant language extensions without some kind of automated check in the CI.

PR #356, opened 2023-12-09
favor `ArchAddrWidth` over `RegAddrWidth (ArchReg ...)`

Ptival:
This is just a suggestion, but we're frequently using `RegAddrWidth (ArchReg arch)` when we also have a type synonym for exactly this, namely `ArchAddrWidth arch`.

I feel like I almost never need to dig into the details of it, so I'd personally prefer only ever seeing `ArchAddrWidth`, but I'm open to this suggestion being rejected for reasons I'm not aware of.


macaw-semmc/src/Data/Macaw/SemMC/Generator.hs
```diff
@@ -102,7 +102,7 @@ data GenState arch ids s =
 
 emptyPreBlock :: RegState (ArchReg arch) (Value arch ids)
               -> Word64
-              -> MM.MemSegmentOff (RegAddrWidth (ArchReg arch))
+              -> MM.MemSegmentOff (ArchAddrWidth arch)
```
RyanGlScott, line 105: `MemSegmentOff (ArchAddrWidth arch)` can be further abbreviated to `ArchSegmentOff arch`. (Similarly for other occurrences of `MemSegmentOff (ArchAddrWidth arch)` in this patch.)

RyanGlScott (approved):

macaw-semmc/src/Data/Macaw/SemMC/Generator.hs
Ptival: I replaced some of those too now.

PR #367, opened 2024-01-23
x86-symbolic: Create an exception type for missing semantics

langston-barrett:
This is an exception that consumers might want to catch.

### DIFF
--- a/x86_symbolic/src/Data/Macaw/X86/Crucible.hs
+++ b/x86_symbolic/src/Data/Macaw/X86/Crucible.hs
@@ -19,6 +19,7 @@ module Data.Macaw.X86.Crucible
     SymFuns(..), newSymFuns
 
     -- * Instruction interpretation
+  , MissingSemantics(..)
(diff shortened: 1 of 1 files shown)


RyanGlScott (approved):
I like it. This brings `macaw-x86-symbolic` more in line with how exceptions are handled in `macaw-aarch32-symbolic` (see [`AArch32Exception`](https://github.com/GaloisInc/macaw/blob/e1d0846be723c02654840808588addf28b270807/macaw-aarch32-symbolic/src/Data/Macaw/AArch32/Symbolic/Functions.hs#L47-L51)).

There's perhaps more that could be done here. For instance, there are other functions in `Data.Macaw.X86.Crucible` that call `error` instead of throwing a custom exception type, and the `termSemantics` function in `macaw-ppc-symbolic` also suffers from a similar problem. I'm happy to defer the task of converting the rest of `macaw` to using structured exceptions to a separate issue, however.

PR #380, opened 2024-04-10
`macaw-base`: Fix `R_X86_64_RELATIVE` offset computation

### LINKED ISSUE #316: `R_X86_64_RELATIVE` relocation offsets are computed incorrectly
`macaw` computes the `relocationOffset` for an `R_ARM_RELATIVE` relocation very carefully:

https://github.com/GaloisInc/macaw/blob/07b693a30905360c4c7d3ffe3b8ccad13ccc191a/base/src/Data/Macaw/Memory/ElfLoader.hs#L669-L676

This is to ensure that you can simply add the `relocationOffset` to the base address to obtain the relocation address without any additional footwork.

On the other hand, `macaw` does _not_ do the same thing on the x86_64 side for `R_X86_64_RELATIVE`:

https://github.com/GaloisInc/macaw/blob/07b693a30905360c4c7d3ffe3b8ccad13ccc191a/base/src/Data/Macaw/Memory/ElfLoader.hs#L548-L558

Instead, it simply slaps on the `addend` without subtracting the `linktimeAddr`. I'm almost positive that this will lead to unexpected behavior if you try to treat `R_X86_64_RELATIVE` the same way that one would treat `R_ARM_RELATIVE`.

RyanGlScott:
Fixes #316.

### DIFF
--- a/base/src/Data/Macaw/Memory/ElfLoader.hs
+++ b/base/src/Data/Macaw/Memory/ElfLoader.hs
@@ -547,10 +547,23 @@ relaTargetX86_64 _ symtab rel addend _isRel =
                         , relocationJumpSlot = True
                         }
     Elf.R_X86_64_RELATIVE -> do
+      -- This relocation has the value B + A where
+      -- - A is the addend for the relocation, and
+      -- - B resolves to the difference between the
+      --   address at which the segment defining the symbol was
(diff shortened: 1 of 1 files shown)


sauclovian-g (approved):
This is consistent with another pile of code for handling these, FWIW.

PR #381, opened 2024-05-09
Fix refinement panic with parallel CFG edges

staslyakhov:
We encountered a binary that caused a panic due to `insertLabeledEdge` returning `Nothing`.
It seems like this could happen in one of two cases:
- The source or the destination vertex doesn't exist
- An edge between source and destination already exists

By using `replaceLabeledEdge` instead, we can avoid panicking in the second case where refinement execution can proceed without issue.
However, it would be nice to make sure that the appearance of a parallel edge isn't an indicator of a larger issue in the algorithm, or if using `replaceLabeledEdge` is a good enough fix.

### DIFF
--- a/refinement/src/Data/Macaw/Refinement/Path.hs
+++ b/refinement/src/Data/Macaw/Refinement/Path.hs
@@ -166,10 +166,13 @@ mkPartialCFG fi = computeBackEdges graph
           tgtId = case M.lookup (FBU.blockID tgtBlock) addrNodes of
                     Just tgtId' -> tgtId'
                     Nothing -> panic $ "Cannot find target ID: " ++ show (FBU.blockID tgtBlock)
+
+          -- Since `tgtId` and `srcId` are confirmed to be valid nodes in the cfg,
+          -- the `Nothing` case implies that the edge already exists, so we don't need to insert it.
+          -- This can occur with jump tables that have common targets (e.g. a `default` case in a switch).
           g = case G.insertLabeledEdge g0 srcId tgtId () of
                 Just (_e, g') -> g'
-                Nothing -> panic $ "Cannot find edge between "
-                                ++ show srcId ++ " and " ++ show tgtId
+                Nothing -> g0
       in g
     buildCFG gr (_addr, pb) =
       let nodeId = case M.lookup (FBU.blockID pb) (cfgAddrNodes gr) of


refinement/src/Data/Macaw/Refinement/Path.hs
```diff
@@ -166,7 +166,7 @@ mkPartialCFG fi = computeBackEdges graph
           tgtId = case M.lookup (FBU.blockID tgtBlock) addrNodes of
                     Just tgtId' -> tgtId'
                     Nothing -> panic $ "Cannot find target ID: " ++ show (FBU.blockID tgtBlock)
-          g = case G.insertLabeledEdge g0 srcId tgtId () of
+          g = case G.replaceLabeledEdge g0 srcId tgtId () of
                 Just (_e, g') -> g'
                 Nothing -> panic $ "Cannot find edge between "
                                 ++ show srcId ++ " and " ++ show tgtId
```
kquick, line 172: You didn't mention which panic was encountered (i.e. the panic message), so if I assume that it was the panic here on line 171, then it could not have been a missing vertex and must therefore have been a duplicate edge.

Since you're here, could you please modify the message here from "Cannot find edge" to "Duplicate edge", which is a better description of this failure.

While your solution works, it's inefficient: the effect of `replaceLabeledEdge` is basically to just update the Edge Label, but since there is no edge labeling in the graph, `g == g0` both before and after the `replaceLabeledEdge`.

At a higher level, if this truly does represent a duplicate edge, I'm a bit concerned because I would not expect that from our CFG representation.  I think it would be useful to add the `MDS.pblockTermStmt pb` information to the trio of panics here in `addEdge` to find out where this duplicate might be coming from.  Based on what that shows, we can figure out if we want to allow duplicate edges, allow only in specific circumstances, or this is indicative of an error elsewhere.

kquick (changes requested):
I think this deserves a little more research: this is pretty unexpected and I'd like to know more about why this happened.

refinement/src/Data/Macaw/Refinement/Path.hs
kquick: Also note that the *implementation* of `replaceLabeledEdge` for the haggle PatriciaTree is to first delete all edges between those nodes and then add the requested edge, which is pretty inefficient; if we decide to allow the duplicate edge, we may just want to return `g0` when that is encountered.

staslyakhov: > You didn't mention which panic was encountered (i.e. the panic message), so if I assume that it was the panic here on line 171, then it could not have been a missing vertex and must therefore have been a duplicate edge.

Yes that's right.

> Since you're here, could you please modify the message here from "Cannot find edge" to "Duplicate edge", which is a better description of this failure.

I'm not sure what the goal of the panic should actually be here. If someone were to call `addEdge` with an invalid `srcId`, then `insertLabeledEdge` would also return `Nothing` and saying that it's a duplicate edge in the panic would be misleading. If that's a valid concern, I see two cases:
1. Having a parallel edge is actually an issue that deserves to be handled via panic. Then we might want to distinguish between this case and the missing vertices case to not have a misleading panic
2. There is no issue with having a parallel edge (I'm guessing if we try to add a duplicate edge it would be turned into a noop). In that case having the panic say "srcId is an invalid vertex" makes sense

I haven't looked very deeply into PatriciaTree and how efficient the operations are, but if that's a concern we can definitely do better than `replaceLabeledEdge` (we'd probably just return `g0` as you suggested). The only possible concern is distinguishing between two possible failure modes of `insertLabeledEdge`.

Since it isn't clear that parallel edges should come up, I'll take a closer look to if what I'm getting in my sample binary makes sense. Thanks for the tip to print out `MDS.pblockTermStmt pb`!

kquick: > I'm not sure what the goal of the panic should actually be here. If someone were to call addEdge with an invalid srcId, then insertLabeledEdge would also return Nothing and saying that it's a duplicate edge in the panic would be misleading.

The code at line 175 looks up `srcId` and has its own panic if it cannot find it.  The map it looks it up from is built at the same time in generated the base graph, so if `srcId` doesn't exist in the graph that's a significant bug (all these `panics` represent bugs in the form of conditions we never expect to happen).  Because both `srcId` and `tgtId` are looked up from this map, it should be essentially impossible for `insertLabeledEdge` to not find the nodes, therefore it's failure should be the duplicate edge insertion case (thus my recommendation for the change in the panic message).

staslyakhov: I did a bit of reversing and I think I have a couples examples of what can happen:

The sample binary (from coreutils) runs into the [`fnmatch_pattern_has_wildcards`](https://www.gnu.org/software/gnulib/coverage/gllib/exclude.c.gcov.html) function. On line 117, it defines `cset` indicating the special characters to search through in `strcspn`. Looks like in the assembly the characters are checked using a jump table that almost always jumps to the same address, but in special cases (when it encounters the characters in `cset`) it jumps somewhere else.

Another example where it panics is in a [jump table](https://github.com/coreutils/coreutils/blob/master/src/du.c#L749) in `du`, where the majority of jump targets are the `otherwise` case.

Now looking into the refinement algorithm, when we encounter a `ParsedLookupTable` we fold over all target addresses in the table and try to call `addEdge` on each one. If there are any duplicate targets, it would seem we would get this panic.

kquick: Now that you've discovered this (thanks for digging in!) this sounds like an entirely reasonable case.  

My recommendation would be to just stick with the `insertLabeledEdge` and change the `Nothing` case to a comment describing that "jump tables with common targets can cause these duplicate edges, and that previous code validated the nodes, so this is a benign case" and return `g0`.  Just a suggestion: feel free to try other approaches, but definitely please document the possibility and scenario for multiple parallel edges for future travelers.

And refinement has been minimally verified, so it's not incredibly surprising there are issues.  It's optional because consulting the solver can be a pretty costly enterprise, so I'll bet most client code doesn't even enable refinement; we should probably change that if we dare.

staslyakhov: I just rewrote the commit to directly handle the `Nothing` case and added a comment to document what happened. Let me know if I can improve anything!

This came up as part of a work to make Reopt use refinement to see how it would improve its jump table recognition capabilities. So far it seems like the solver definitely helps, but it is a little expensive!

kquick (approved):
Thanks!

PR #382, opened 2024-05-20
parse subroutine type declaration formal parameters

Ptival:
This is useful for Reopt as we want to get more accurate type information for library functions.

Is this acceptable for other clients?

### DIFF
--- a/base/ChangeLog.md
+++ b/base/ChangeLog.md
@@ -32,3 +32,6 @@
 - The `Hashable` and `HashableF` instances for `App f` now require
   `TestEquality f` constraints. (This is needed to support `hashable-1.4.*`,
   which adds `Eq` as a superclass to `Hashable`.)
+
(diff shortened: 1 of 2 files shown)


RyanGlScott:
> Is this acceptable for other clients?

Hard to say, since I don't know if there _are_ any other clients of this code besides `reopt`. (A GitHub search didn't turn up anything, but that's not conclusive.) My gut reaction is that using a `Variable` here would be more precise (and arguably better) than using a `DIE` in most cases, but it's hard to anticipate if there are other use cases where a `DIE` would be preferable.

This is all to say: I am inclined to make this change and see if anybody complains. (If they do, we can always add the `DIE`s back alongside the `Variable`s.) It would be worth advertising this API change in the [`macaw-base` changelog](https://github.com/GaloisInc/macaw/blob/master/base/ChangeLog.md) as a precaution, however.

PR #390, opened 2024-07-03
Simple CLI for macaw-symbolic

### LINKED ISSUE #389: Simple CLI frontend for `macaw-*-symbolic`
Ocassionally, I have a question about `macaw-*-symbolic`'s semantics. The only way I can verify my hypotheses is to run a higher-level tool built on top of Macaw. However, such tools generally have a considerable amount of code related to features other than simply executing machine code, obfuscating the results (e.g., is the problem in Macaw, or in the higher-level tool?).

It would be great if the various `macaw-*-symbolic` packages had simple command-line interfaces that would just load binaries and start at `main` with a sensible initial memory and register state. Ideally, these could also support `macaw-syntax` S-expression programs, a la `crucible-cli`. We'd likely want to make a generic library (`macaw-cli`), with instantiations for each particular architecture.

langston-barrett:
Fixes #389.

### DIFF
--- a/macaw-cli/src/Data/Macaw/CLI.hs
+++ b/macaw-cli/src/Data/Macaw/CLI.hs
@@ -17,8 +17,8 @@ module Data.Macaw.CLI
 import Control.Lens qualified as Lens
 import Data.ByteString.Char8 qualified as BS8
 import Data.List qualified as List
-import Data.Map qualified as Map 
-import Data.Text qualified as Text 
+import Data.Map qualified as Map
+import Data.Text qualified as Text
 import GHC.TypeLits (KnownNat)
 
 -- First-party
@@ -69,7 +69,7 @@ sim archInfo archVals pltStubInfo extractor elfHeaderInfo opts = do
   let discState = MST.binaryDiscState (MST.mainBinaryInfo binfo)
   let funInfos = Map.elems (discState Lens.^. MD.funInfo)
   let entryFn8 = BS8.pack (Text.unpack entryFn)
-  let isEntry sdfi = 
+  let isEntry sdfi =
         case sdfi of
           Some.Some dfi ->
             case MD.discoveredFunSymbol dfi of
@@ -95,6 +95,6 @@ ppSimRes =
     MST.SimulationAborted -> "Aborted!"
     MST.SimulationTimeout -> "Timeout!"
     MST.SimulationPartial -> "Partial!"  -- TODO: What does this mean?
-    MST.SimulationResult MST.Unsat -> "Always returns 0"
-    MST.SimulationResult MST.Sat -> "May return non-zero"
+    MST.SimulationResult MST.Unsat -> "May return non-zero"
+    MST.SimulationResult MST.Sat -> "Always returns 0"
     MST.SimulationResult MST.Unknown -> "Solver returned unknown!"
--- a/macaw-cli/src/Data/Macaw/CLI.hs
+++ b/macaw-cli/src/Data/Macaw/CLI.hs
@@ -1,0 +1,85 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE ImplicitParams #-}
+{-# LANGUAGE ImportQualifiedPost #-}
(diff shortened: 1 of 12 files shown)


macaw-cli/src/Data/Macaw/CLI.hs
```diff
@@ -0,0 +1,100 @@
...
+        Just <$> MST.simulateAndVerify solver WS.defaultLogData bak execFeatures archInfo archVals binfo mmPreset extractor dfi
+
+ppSimRes :: MST.SimulationResult -> Text.Text
+ppSimRes =
+  \case
+    MST.SimulationAborted -> "Aborted!"
+    MST.SimulationTimeout -> "Timeout!"
+    MST.SimulationPartial -> "Partial!"  -- TODO: What does this mean?
+    MST.SimulationResult MST.Unsat -> "Always returns 0"
+    MST.SimulationResult MST.Sat -> "May return non-zero"
```
RyanGlScott, line 99: These two lines are reversed, I believe. Per [this comment](-nostdlib -static -fno-stack-protector) in `Data.Macaw.Symbolic.Testing`, the program succeeds (i.e., checking the negated goal returns `Unsat`) if the return value is non-zero, and the program fails (i.e., checking the negated goal returns `Sat`) if the return value is zero.

macaw-cli/src/Data/Macaw/CLI.hs
```diff
@@ -0,0 +1,100 @@
...
+
+-- First-party
+import Data.ElfEdit qualified as Elf
+import Data.Macaw.Architecture.Info qualified as MAI
+import Data.Macaw.CFG qualified as MCFG
+import Data.Macaw.CLI.Options qualified as MCO
+import Data.Macaw.Discovery qualified as MD
+import Data.Macaw.Memory.ElfLoader.PLTStubs qualified as MPLT
+import Data.Macaw.Symbolic qualified as MS
+import Data.Macaw.Symbolic.Testing qualified as MST
```
RyanGlScott, line 32: We are now using `Data.Macaw.Symbolic.Testing` for reasons beyond just simple testing. I wonder if we should consider migrating the name to something more appropriate. (`Data.Macaw.Symbolic.Driver`, perhaps?)

macaw-cli/src/Data/Macaw/CLI.hs
RyanGlScott: Fixed in https://github.com/GaloisInc/macaw/pull/390/commits/248518df082ff4999009fe24314e4842e4cdd51d.

macaw-cli/src/Data/Macaw/CLI.hs
RyanGlScott: I've also observed that `Data.Macaw.Symbolic.Testing` has some peculiar conventions that we may not want to use in `macaw-cli`. For instance, `Data.Macaw.Symbolic.Testing` won't take assumptions into account when checking an entrypoint function [unless it begins with the prefix "`test_and_verify_`"](https://github.com/GaloisInc/macaw/blob/645754178a034c22e97aac49a962c99b387aaa01/symbolic/src/Data/Macaw/Symbolic/Testing.hs#L433), so most functions won't check if they use the memory model in valid ways. This seems undesirable for `macaw-cli`'s needs.

Perhaps we could factor out the code used in common between `Data.Macaw.Symbolic.Testing` and the new module we use to power `macaw-cli` (be it named `Data.Macaw.Symbolic.Driver` or something else).

RyanGlScott:
@langston-barrett, what are your plans for this PR now that https://github.com/GaloisInc/macaw/pull/423 has landed? https://github.com/GaloisInc/macaw/pull/423 only adds support for S-expression programs, whereas this PR is about binaries. Do you want to extend `macaw-x86-cli` to support binaries as well?

langston-barrett:
> @langston-barrett, what are your plans for this PR now that https://github.com/GaloisInc/macaw/pull/423 has landed?

I don't have further plans for this PR; as we've discussed in other forums, we should likely work towards a CLI that ingests binaries, but the testing harness code I'ver reused here is actually not reusable enough to suit this purpose. Instead, we should do a larger refactor of Macaw's APIs to make it easier to create frontends, and spend more effort creating one that has all of the conventions and functionality we actually think it should.

PR #399, opened 2024-07-16
Use Crucible goal-proving helpers in testing code

### LINKED ISSUE #391: `symbolic`: Use new Crucible goal-proving helpers
Soon, Crucible will have [a new API for dispatching proof obligations](https://github.com/GaloisInc/crucible/pull/1216). Uses of that API could replace the following code in `macaw-symbolic`:

https://github.com/GaloisInc/macaw/blob/645754178a034c22e97aac49a962c99b387aaa01/symbolic/src/Data/Macaw/Symbolic/Testing.hs#L318-L352

langston-barrett:
Fixes #391


RyanGlScott:
This looks reasonable at a glance, although I think you'll need to bump the `crucible` submodule in order for it to work.

RyanGlScott (approved):

Issue #387, opened 2024-07-01
`macaw-ppc-symbolic`: Allow simulating syscalls


RyanGlScott:
Currently, `macaw-symbolic`'s [`lookupSyscallHandle`](https://github.com/GaloisInc/macaw/blob/645754178a034c22e97aac49a962c99b387aaa01/symbolic/src/Data/Macaw/Symbolic.hs#L1126-L1128) memory model operation is wired up to support x86-64 and AArch32, but not PowerPC. We should add support for PowerPC, following the blueprint laid out in https://github.com/GaloisInc/macaw/pull/246.

Note that PPC32 and PPC64 have ever-so-slightly different calling conventions for syscalls, as seen in `man 2 syscall`:

```
       Arch/ABI      arg1  arg2  arg3  arg4  arg5  arg6  arg7  Notes
       ──────────────────────────────────────────────────────────────
       powerpc       r3    r4    r5    r6    r7    r8    r9
       powerpc64     r3    r4    r5    r6    r7    r8    -
```
```
       Arch/ABI    Instruction           System  Ret  Ret  Error    Notes
                                         call #  val  val2
       ───────────────────────────────────────────────────────────────────
       powerpc     sc                    r0      r3   -    r0       1
       powerpc64   sc                    r0      r3   -    cr0.SO   1

       Notes:

       [1] On  a few architectures, a register is used as a boolean (0 indicating no error,
           and -1 indicating an error) to signal that the system call failed.   The  actual
           error  value is still contained in the return register.  On sparc, the carry bit
           (csr) in the processor status register (psr) is used instead of a full register.
           On powerpc64, the summary overflow bit (SO) in field 0 of the condition register
           (cr0) is used.
```

Still, these differences are minor enough that we may be able to simply put all of the PowerPC-related syscall information into a single `PPCSyscall` constructor in `PPCPrimFn` and do slightly different things with this information depending on whether a PowerPC binary uses 32 bits or 64 bits.

sauclovian-g:
FWIW, the carry flag is traditional and what's used on almost all older architectures. (Also, when the failure goes in a full register it's not necessarily -1, just nonzero. On mips you'll usually get 1.)

not immediately relevant of course

Issue #385, opened 2024-06-21
`macaw-ppc-symbolic`: Export `SemanticsError`


RyanGlScott:
`macaw-ppc-symbolic` currently does not export the [`SemanticsError`](https://github.com/GaloisInc/macaw/blob/645754178a034c22e97aac49a962c99b387aaa01/macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs#L60) exception type. It would be nice to export this so that these exceptions can be caught in downstream applications, much like we can do for x86_64 and AArch32 exception types (see https://github.com/GaloisInc/macaw/pull/370).

PR #370, opened 2024-01-24
{aarch32,x86}-symbolic: A bit more work on exceptions

langston-barrett:
Fixes #369.

PR #401, opened 2024-07-19
`macaw-ppc-symbolic`: Add syscall support, export `SemanticsError`

RyanGlScott:
This PR contains a collection of commits needed to bring `macaw-ppc-symbolic`'s handling of system calls and semantics-related exceptions in line with `macaw-x86-symbolic` and `macaw-aarch32-symbolic`:

## `macaw-ppc-symbolic`: Export `SemanticsError`

Fixes https://github.com/GaloisInc/macaw/issues/385 (and finishes the work started in https://github.com/GaloisInc/macaw/pull/370).

## `macaw-ppc-symbolic`: Add support for simulating syscalls

This adds the necessary changes to `macaw-ppc-symbolic` and `macaw-ppc` in order to simulate system calls, similarly to how it is done when simulating x86-64 and AArch32 code:

* In `macaw-ppc`, remove `PPCSyscall` from `TermStmt` and instead make it a constructor for `PPCPrimFn`. (Note that there are some minor discrepancies between which registers are used in PPC32 versus PPC64, which we explain in the Haddocks for the new `PPCSyscall` constructor.)
* Update `macaw-ppc`'s `ppcInstructionMatcher` function so that calls to the `sc` (system call) instruction make use of `PPCSyscall`.
* Update `macaw-ppc-symbolic`'s `ppcGenFn` function to make it possible to hook into PPC system calls using `MacawLookupSyscallHandle`.

Fixes https://github.com/GaloisInc/macaw/issues/387.

### DIFF
--- a/macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs
+++ b/macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs
@@ -34,7 +34,6 @@ import qualified Lang.Crucible.Simulator.RegMap as C
 import qualified Lang.Crucible.Simulator.SimError as C
 import qualified Lang.Crucible.Types as C
 import qualified What4.Interface as C
-import qualified What4.InterpretedFloatingPoint as C
 
 import qualified SemMC.Architecture.PPC as SP
 import qualified SemMC.Util as U
--- a/macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs
+++ b/macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs
@@ -42,6 +42,7 @@ import qualified Data.Macaw.Symbolic as MS
 import qualified Data.Macaw.PPC as MP
 
 import qualified Data.Macaw.PPC.Symbolic.AtomWrapper as A
+import qualified Data.Macaw.PPC.Symbolic.Panic as P
 
 data SomeSymFun sym where
   SomeSymFun :: Ctx.Assignment C.BaseTypeRepr ps -> C.BaseTypeRepr r -> C.SymFn sym ps r -> SomeSymFun sym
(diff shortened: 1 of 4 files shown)


macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs
```diff
@@ -314,6 +314,8 @@ funcSemantics sf pf s =
       fval <- lookupApplySymFun sym sf ("vec_" ++ name) C.knownRepr args C.knownRepr
       ptrVal <- LL.llvmPointer_bv sym fval
       return (ptrVal, s)
+    MP.PPCSyscall {} ->
+      P.panic P.PPC "funcSemantics" ["The PPC syscall primitive should be eliminated and replaced by a handle lookup"]
```
langston-barrett, line 318: Should we just remove this constructor? Or is it necessary for this to happen after translation into a Macaw-PPC CFG?

macaw-ppc/src/Data/Macaw/PPC/Arch.hs
```diff
@@ -755,7 +825,44 @@ ppcInstructionMatcher :: forall var ids s n
...
+      let r7 = regs ^. MC.boundValue (PPC_GP (D.GPR 7))
+      let r8 = regs ^. MC.boundValue (PPC_GP (D.GPR 8))
+      let r9 = regs ^. MC.boundValue (PPC_GP (D.GPR 9))
+      let sc = PPCSyscall w r0 r3 r4 r5 r6 r7 r8 r9
+      res <- G.addExpr =<< evalArchFn sc
+
+      resVal       <- G.addExpr (G.AppExpr (MC.TupleField knownRepr res PL.index0))
+      errorCondVal <- G.addExpr (G.AppExpr (MC.TupleField knownRepr res PL.index1))
+
+      G.setRegVal (PPC_GP (D.GPR 3)) resVal
```
langston-barrett, line 845: Perhaps something like:
```suggestion
      -- See the documentation on 'PPCSyscall' for a description of the registers used to return values from syscalls
      G.setRegVal (PPC_GP (D.GPR 3)) resVal
```

langston-barrett (approved):
I'm not deeply familiar with the surrounding code, but this looks generally reasonable to me!

macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs
RyanGlScott: It's still necessary to have the `PPCSyscall` primitive function, as it exists for some (but not all) parts of the `macaw-symbolic` translation pipeline. See [`Note [Syscalls]`](https://github.com/GaloisInc/macaw/blob/83d3907054181b17be365bdea9abb25edd48aae3/x86_symbolic/src/Data/Macaw/X86/Symbolic.hs#L396-L424).

PR #403, opened 2024-07-23
Improve handling of JumpBound truncation constraints

staslyakhov:
# Background
Given some symbolic value, `x`, we'd like to compute its possible upper and lower bounds after it is truncated to `w` bits.

To do this, we first find the bound of x by (recursively) calling `exprRangePred`. This bound is a statement of the following form (see `RangePred` for more info): "`r` bits of `x` are bounded between `low` and `high`".

Then, we check the following:
- If `x` has a bound `(r, l, h)` AND
- If `r` is less than or equal to `w` => Pass-through the bound `(r, l, h)`

Otherwise, we deem `x` "unbounded"

# Proposal
Declaring `x` unbounded in the second case seems to throw away useful information that causes many jump tables to remain unclassified. We attempt to improve on that in this commit.

Consider an example where `x` is bounded by `(64, 0, 10)` (that is, the 64 bits of `x` are constrained to be between 0 and 10) and we want to find the bound of truncating `x` to 8 bits.

With the current logic, since 64 > 8, we'd declare x unbounded. However, the bound `(8, 0, 10)` should also be valid: if 64 bits of `x` are bounded to [0, 10], then surely 8 bits of `x` also lie between 0 and 10.

If the upper bound is instead larger than the largest 8-bit value, we can truncate it to the largest value.
For example, `(64, 0, 10000)` becomes `(8, 0, 255)`.
Instead of losing the bound completely, we're able to tighten it!

# Evaluation
Using Reopt's ground truth evaluation, we found that this change significantly improved jump table classification in [coreutil binaries](https://github.com/junxzm1990/x86-sok/tree/master/testsuite/linux/utils) compiled with `GCC_O1`. Out of 912 total jump tables in the dataset, macaw (launched through Reopt) currently reaches + classifies 109 jump tables. After this change, macaw correctly classifies 857/912 jump tables.

### DIFF
--- a/base/src/Data/Macaw/AbsDomain/JumpBounds.hs
+++ b/base/src/Data/Macaw/AbsDomain/JumpBounds.hs
@@ -386,8 +386,21 @@ exprRangePred info e = do
             -- Compare the range constraint with the output number of bits.
             -- If the bound on r covers at most the truncated number
             -- of bits, then we just pass it through.
-          , Just LeqProof <- testLeq (rangeWidth r) w ->
-            SomeRangePred r
+          -> case testLeq (rangeWidth r) w of
+              Just LeqProof -> SomeRangePred r
+
+              -- Otherwise, we need to rewrite the constraints to be on w bits.
+              -- We can do that by clamping the lower/upper bounds if they're
+              -- outside of the range [0, 2^w-1].
+              Nothing ->
+                let
+                  truncUnsigned :: Natural -> Natural
+                  truncUnsigned = fromInteger . unsignedClamp w . toInteger
+
+                  lowerBound = truncUnsigned $ rangeLowerBound r
+                  upperBound = truncUnsigned $ rangeUpperBound r
+                in
+                  SomeRangePred (mkRangeBound w lowerBound upperBound) 
         _ -> NoRangePred
 
 -- | This returns a natural number with a computed upper bound for the


base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -386,8 +386,18 @@ exprRangePred info e = do
...
-          , Just LeqProof <- testLeq (rangeWidth r) w ->
-            SomeRangePred r
+          -> case testLeq (rangeWidth r) w of
+              Just LeqProof -> SomeRangePred r
+              
+              -- Otherwise, our bound remains valid for the truncated number
+              -- of bits and we can tighten the bounds to at most @2^w-1@ (inclusive)
+              Nothing ->
+                let 
+                  truncUnsigned = fromInteger . unsignedClamp w . toInteger
```
RyanGlScott, line 396: Can you add an explicit `Natural -> Natural` type signature to `truncUnsigned`? The `fromInteger`/`toInteger` functions work over many types, so without a type signature, you'd have to look elsewhere to realize what type it is actually being used on.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -386,8 +386,18 @@ exprRangePred info e = do
             -- Compare the range constraint with the output number of bits.
             -- If the bound on r covers at most the truncated number
             -- of bits, then we just pass it through.
-          , Just LeqProof <- testLeq (rangeWidth r) w ->
-            SomeRangePred r
+          -> case testLeq (rangeWidth r) w of
+              Just LeqProof -> SomeRangePred r
+              
+              -- Otherwise, our bound remains valid for the truncated number
+              -- of bits and we can tighten the bounds to at most @2^w-1@ (inclusive)
```
RyanGlScott, line 393: Worth mentioning in this comment that we may need to clamp the lower or upper bounds if the fall outside of the range `[0, 2^w-1]`.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -386,8 +386,18 @@ exprRangePred info e = do
...
+              Nothing ->
+                let 
+                  truncUnsigned = fromInteger . unsignedClamp w . toInteger
+                  lowerBound = truncUnsigned $ rangeLowerBound r
```
RyanGlScott, line 397: Actually, come to think of it, can the lower bound ever be lower than `0`? The type of `rangeLowerBound` is a `Natural`, after all, not an `Integer`.

As such, I think we could get away with not clamping the lower bounds here, which would make this part of the code slightly more efficient.

staslyakhov: I suppose one interesting edge case here would be if `lower_bound > 2^w - 1`. Then the clamping would tighten the bound to be `2^w -1`, right?

I wasn't too concerned with efficiency here, so if this is the only edge case maybe it should be handled differently?

RyanGlScott: Ah, fair enough. In that case, I suppose we really do have to clamp the lower bound to be on the safe side.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
staslyakhov: I think what I meant with "tighten the bounds" is too ambiguous.  How about the following:

> Otherwise, we need to rewrite the constraints to be on `w` bits. We can do that by clamping the lower/upper bounds if they're outside of the range `[0, 2^w-1]`.

RyanGlScott: That sounds reasonable to me.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
staslyakhov: Done! I rewrote the commit.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
staslyakhov: Replaced with this comment.

RyanGlScott (approved):
Thanks!

PR #405, opened 2024-07-23
add a `Bits` instance to `MemInt`

Ptival:
Note: this made me realize that `addrBitSize` is a bit ill-named as I don't think it has anything to do with addresses.

### DIFF
--- a/base/src/Data/Macaw/Memory.hs
+++ b/base/src/Data/Macaw/Memory.hs
@@ -465,6 +465,20 @@ instance MemWidth w => Integral (MemInt w) where
     where (q,r) = memIntValue x `quotRem` memIntValue y
   toInteger = toInteger . memIntValue
 
+instance MemWidth w => Bits (MemInt w) where
+  MemInt x .&. MemInt y = memInt (x .&. y)
+  MemInt x .|. MemInt y = memInt (x .|. y)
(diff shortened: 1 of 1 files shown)


base/src/Data/Macaw/Memory.hs
```diff
@@ -465,6 +465,20 @@ instance MemWidth w => Integral (MemInt w) where
     where (q,r) = memIntValue x `quotRem` memIntValue y
   toInteger = toInteger . memIntValue
 
+instance MemWidth w => Bits (MemInt w) where
+  MemInt x .&. MemInt y = memInt (x .&. y)
+  MemInt x .|. MemInt y = memInt (x .|. y)
+  MemInt x `xor` MemInt y = memInt (x `xor` y)
+  complement (MemInt x) = memInt (complement x)
+  MemInt x `shift` i = memInt (x `shift` i)
+  MemInt x `rotate` i = MemInt (x `rotate` i)
```
RyanGlScott, line 474: Should this use the `memInt` smart constructor for consistency with the other definitions?

RyanGlScott (approved):
LGTM, aside from one question.

base/src/Data/Macaw/Memory.hs
Ptival: Indeed that'd be better. Will fix.

PR #406, opened 2024-07-23
support alignment ops when processing statements in `RegisterUse`

Ptival:
For some reason we were very conservative in our support or abstract
operations over the processor state in the `RegisterUse` analysis.

In particular, we were failing to process code such as:

r23 := (bv_and r21 (0xfffffffffffffff0 :: [64]))

whose goal is to align the value in r21 at a 16-byte boundary.

This resulted in us failing to analyze some code that was realigning its
stack pointer.  With this change, the same code succeeds at propagating
the abstract stack pointer offset forward.

### DIFF
--- a/base/src/Data/Macaw/Analysis/RegisterUse.hs
+++ b/base/src/Data/Macaw/Analysis/RegisterUse.hs
@@ -71,6 +71,7 @@ import           Control.Monad (unless, when, zipWithM_)
 import           Control.Monad.Except (MonadError(..), Except)
 import           Control.Monad.Reader (MonadReader(..), ReaderT(..), asks)
 import           Control.Monad.State.Strict (MonadState(..), State, StateT, execStateT, evalState, gets, modify)
+import qualified Data.Bits as Bits
 import qualified Data.ByteString.Char8 as BSC
 import qualified Data.ByteString as BS
 import           Data.Foldable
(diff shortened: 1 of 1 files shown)


base/src/Data/Macaw/Analysis/RegisterUse.hs
```diff
@@ -678,6 +679,8 @@ processApp aid app = do
       setAssignVal aid (FrameExpr (o+fromInteger c))
     BVSub _ (FrameExpr o) (IVCValue (BVCValue _ c)) ->
       setAssignVal aid (FrameExpr (o-fromInteger c))
+    BVAnd _ (FrameExpr o) (IVCValue (BVCValue _ c)) ->
+      setAssignVal aid (FrameExpr (o Bits..&. memInt (fromInteger c)))
```
RyanGlScott, line 683: For consistency with the code above, you could also write this as:

```suggestion
      setAssignVal aid (FrameExpr (o Bits..&. fromInteger c))
```

base/src/Data/Macaw/Analysis/RegisterUse.hs
```diff
@@ -678,6 +679,8 @@ processApp aid app = do
       setAssignVal aid (FrameExpr (o+fromInteger c))
     BVSub _ (FrameExpr o) (IVCValue (BVCValue _ c)) ->
       setAssignVal aid (FrameExpr (o-fromInteger c))
+    BVAnd _ (FrameExpr o) (IVCValue (BVCValue _ c)) ->
```
RyanGlScott, line 682: Does it also make sense to add a `BVAnd _ (IVCValue (BVCValue _ c)) (FrameExpr o)` case for symmetry with the two `BVAdd` cases above? (I'm not sure if this case would ever occur in practice, but I did find myself wondering about its omission here.)

RyanGlScott (approved):
This is definitely an improvement.

base/src/Data/Macaw/Analysis/RegisterUse.hs
Ptival: Agreed!

base/src/Data/Macaw/Analysis/RegisterUse.hs
Ptival: I agree that in practice this version is the most likely to be generated by Macaw, but it doesn't hurt to add, so I will.

Ptival:
@RyanGlScott I have addressed your comments, and also added two comments that I fell will help me if I ever need to look at this code again.

Issue #1, opened 2017-12-14
Build error due to -Werror and incomplete pattern match


TomMD:
Incomplete pattern match:

```
Building library for macaw-x86-0.0.1..
[11 of 15] Compiling Data.Macaw.X86.Getters ( src/Data/Macaw/X86/Getters.hs, /Users/tommd/dev/reopt/dist-newstyle/build/x86_64-osx/ghc-8.2.1/macaw-x86-0.0.1/build/Data/Macaw/X86/Getters.o )

src/Data/Macaw/X86/Getters.hs:182:3: warning: [-Wincomplete-patterns]
    Pattern match(es) are non-exhaustive
    In a case alternative: Patterns not matched: (F.Mem256 _)
    |
182 |   case v of
    |   ^^^^^^^^^...
```

PR #411, opened 2024-07-26
Implement `macaw-riscv-symbolic`

### LINKED ISSUE #409: Add `macaw-riscv-symbolic` package
Currently, `macaw-riscv` is the only `macaw` backend that lacks symbolic execution support via Crucible. We should add a `macaw-riscv-symbolic` package for this.

RyanGlScott:
This adds the necessary plumbing to simulate Macaw-lifted RISC-V binaries using `macaw-symbolic`. This proves relatively straightforward, given that RISC-V does not have a lot of special primitive functions or statements to deal with. I have also added a basic test suite to ensure that `macaw-riscv-symbolic` works on end-to-end examples.

Fixes #409.


macaw-riscv-symbolic/src/Data/Macaw/RISCV/Symbolic.hs
```diff
@@ -0,0 +1,299 @@
...
+
+archUpdateReg :: (G.KnownRV rv, Typeable rv)
+              => C.RegEntry sym (MS.ArchRegStruct (MR.RISCV rv))
+              -> MR.RISCVReg rv tp
+              -> C.RegValue sym (MS.ToCrucibleType tp)
+              -> C.RegEntry sym (MS.ArchRegStruct (MR.RISCV rv))
+archUpdateReg regEntry reg val =
+  case updateReg reg (const $ C.RV val) (C.regValue regEntry) of
+    Just r -> regEntry { C.regValue = r }
+    Nothing -> error $ "unexpected register: " ++ show (MC.prettyF reg)
```
kquick, line 202: I would expect this to be a "this should never happen" situation, in which case I would recommend using `panic` which provides more information for debugging.

macaw-riscv-symbolic/src/Data/Macaw/RISCV/Symbolic/Repeat.hs
```diff
@@ -0,0 +1,30 @@
+{-# LANGUAGE DataKinds #-}
```
kquick, line 1: Should we consider moving this to parameterized-utils?

macaw-riscv-symbolic/tests/Main.hs
```diff
@@ -0,0 +1,218 @@
...
+data SaveSMT = SaveSMT Bool
+  deriving (Eq, Show)
+
+instance TTO.IsOption SaveSMT where
+  defaultValue = SaveSMT False
+  parseValue v = SaveSMT <$> TTO.safeReadBool v
+  optionName = pure "save-smt"
+  optionHelp = pure "Save SMT sessions to files in /tmp for debugging"
+
+-- | A tasty option to have the test suite save the macaw IR for each test case to /tmp for debugging purposes
```
kquick, line 62: I'm wondering if these should just save to the current directory rather than `/tmp`.  Dropping files in a global location mixes them with other runs and could lead to looking at the wrong file.  Another thought is that if we want to generate these during CI and possibly save them as artifacts, it will be easier to generate them directly in the right location (e.g. `mkdir test_out; cd test_out; cabal run macaw-riscv-symbolic-tests -- --save-smt --save-macaw` with the `test_out` being designated in the `ci.yaml` `artifacts` clause).

kquick (approved):

macaw-riscv-symbolic/src/Data/Macaw/RISCV/Symbolic.hs
RyanGlScott: It is possible to observe this `error` if you call `archUpdateReg` with RISC-V's `x0` register, which we intentionally do not include in the `regIndexMap` (see the comments for `gprRegs` above). I suppose there is a philosophical question of whether we should:

1. Anticipate this situation and throw an `error` message here, or
2. Document that this function has a precondition that the `RISCVReg` argument must belong to the set of registers that `macaw-riscv-symbolic` tracks, and panic if this precondition is violated

Personally, I lean more towards option (1), and that's how other `macaw-*-symbolic` backends currently implement their own versions of `archUpdateReg`. I could be talked into implementing (2) if you prefer that, however.

macaw-riscv-symbolic/tests/Main.hs
RyanGlScott: This is another piece of code that I copied over from the `macaw-*-symbolic` test suites. I don't have a strong feeling one way or another about where these files live. Given that `--save-smt` isn't enabled by default, I think we could put these files anywhere.

In fact, I think it would be better if these options took an explicit directory as an argument so that the user can control where the files should live. That way, we could just say `cabal run macaw-riscv-symbolic-tests --save-smt test_out --save-macaw test_out` directly, without needing to `cd` to and fro.

Does that sound reasonable? If so, I'll open a separate issue to track this idea.

macaw-riscv-symbolic/src/Data/Macaw/RISCV/Symbolic.hs
kquick: I suppose my concern is that if you do get an "unexpected register" error and it's not caught and handled elsewhere that it's hard to track down and even when you do, the context of what is wrong takes some time to deduce.  I'm not clear from your two choices why #1 would use `error` and #2 would use `panic`.  I'm somewhat mollified by the argument that this fits the trend elsewhere, but I also wonder if the trend was well thought out or was a case of "I guess I need something here so the compiler doesn't complain about a missing case".  I trust your instincts on the best way to handle this though.

macaw-riscv-symbolic/tests/Main.hs
kquick: That sounds reasonable, and that way we can address that concern independently and simultaneously for all similar situations.

macaw-riscv-symbolic/src/Data/Macaw/RISCV/Symbolic/Functions.hs
```diff
@@ -0,0 +1,54 @@
...
+newSymFuns :: forall sym . (C.IsSymInterface sym) => sym -> IO (SymFuns sym)
+newSymFuns _sym = pure SymFuns
+
+termSemantics :: SymFuns sym
+              -> MR.RISCVTermStmt rv (RA.AtomWrapper (C.RegEntry sym))
+              -> S p rv sym rtp bs r ctx
+              -> IO (C.RegValue sym C.UnitType, S p rv sym rtp bs r ctx)
+termSemantics _fs term _s = case term of {}
+
+type S p rv sym rtp bs r ctx = C.CrucibleState p sym (MS.MacawExt (MR.RISCV rv)) rtp bs r ctx
```
Ptival, line 54: Maybe I'd personally move this type alias above `funcSemantics` just for the reader.

Ptival (approved):
I can't follow all details, but overall this all looks reasonable to me.

macaw-riscv-symbolic/src/Data/Macaw/RISCV/Symbolic.hs
RyanGlScott: > I'm not clear from your two choices why https://github.com/GaloisInc/macaw/issues/1 would use `error` and https://github.com/GaloisInc/macaw/issues/2 would use `panic`.

My reasoning is that it's not just a missing case: it's genuinely possible for a `macaw-riscv-symbolic` API user to trigger this error if they are not careful when calling it. Whether that deserves an `error` or `panic` is a matter of taste, I suppose, since although it's not _impossible_ to trigger this error, it's nevertheless very unlikely.

macaw-riscv-symbolic/ChangeLog.md
```diff
@@ -0,0 +1,5 @@
+# Revision history for macaw-riscv-symbolic
+
+## 0.1.0.0 -- YYYY-mm-dd
```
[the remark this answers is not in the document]
RyanGlScott: If this is overly generic, then this (which I copied [this file](https://github.com/GaloisInc/macaw/blob/eb0a3c7ccc86fd8173633d5007381e62348c48cd/macaw-ppc-symbolic/ChangeLog.md) from) is also overly generic :)

Would changing the dates to "TBA" or something similar address your concern here?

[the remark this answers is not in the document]
Ptival: I guess if we're leaving it like this until a potential Hackage release, that's okay.

macaw-riscv-symbolic/tests/Main.hs
RyanGlScott: I've opened https://github.com/GaloisInc/macaw/issues/413 for this.

macaw-riscv-symbolic/src/Data/Macaw/RISCV/Symbolic.hs
RyanGlScott: I've converted these to panics. @kquick has opened https://github.com/GaloisInc/macaw/issues/412 to track the remaining task of changing the return types of `GenArchVals`' `{lookup,update}Reg` functions in `macaw-symbolic`, which are ultimately the reason why these functions have to be partial in the first place.

Issue #349, opened 2023-11-10
symbolic-syntax: Syntactic sugar


langston-barrett:
macaw-symbolic encodes functions as Crucible CFGs that take and return a struct of all the register values. This struct has a lot of fields. Here's a very simple no-op Macaw x86_64 CFG written in the `symbolic-syntax`:
```
(defun @test ((regs (Struct
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      Bool           
                      (Ptr 3)  
                      (Ptr 2)  
                      (Ptr 2)  
                      (Ptr 2)  
                      (Ptr 2)  
                      (Ptr 2)  
                      (Ptr 2)  
                      (Ptr 2)  
                      (Ptr 2)  
                      (Ptr 80) 
                      (Ptr 80) 
                      (Ptr 80) 
                      (Ptr 80) 
                      (Ptr 80) 
                      (Ptr 80) 
                      (Ptr 80) 
                      (Ptr 80) 
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512)
                      (Ptr 512))))
                    (Struct
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
                      (Ptr 64) 
[...]

PR #422, opened 2024-08-15
macaw-x86-syntax: Syntactic sugar for macaw-x86-symbolic CFGs

langston-barrett:
Towards #349


x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
```diff
@@ -124,6 +140,42 @@ rdx = Ctx.extendIndex (Ctx.nextIndex (Ctx.size3 @(MM.LLVMPointerType 64) @(MM.LL
...
+
+r10 :: Ctx.Index (MacawCrucibleRegTypes M.X86_64) (MM.LLVMPointerType 64)
+r10 = Ctx.extendIndex (Ctx.nextIndex (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.size6 @(MM.LLVMPointerType 64) @(MM.LLVMPointerType 64) @(MM.LLVMPointerType 64) @(MM.LLVMPointerType 64) @(MM.LLVMPointerType 64) @(MM.LLVMPointerType 64))))))))
+
+r11 :: Ctx.Index (MacawCrucibleRegTypes M.X86_64) (MM.LLVMPointerType 64)
+r11 = Ctx.extendIndex (Ctx.nextIndex (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.incSize @_ @(MM.LLVMPointerType 64) (Ctx.size6 @(MM.LLVMPointerType 64) @(MM.LLVMPointerType 64) @(MM.LLVMPointerType 64) @(MM.LLVMPointerType 64) @(MM.LLVMPointerType 64) @(MM.LLVMPointerType 64)))))))))
```
RyanGlScott, line 165: I wonder if you could avoid some repetition by defining specialized versions of `incSize` and friends that are specialized to `LLVMPointerType 64`.

langston-barrett: Simplified, let me know what you think!

macaw-x86-syntax/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+This package provides concrete syntax for macaw-x86-symbolic types and
+operations.
+
+Concretely, it implements a `ParserHooks` for use with [`crucible-syntax`][syn].
+This `ParserHooks` supports the following types and operations:
+
+**Types**:
+
+- `X86Regs`: the struct of all x86_64 registers
```
RyanGlScott, line 11: Should `X86Reg` also be defined here?

x86_symbolic/tests/Main.hs
```diff
@@ -70,6 +72,17 @@ ingredients = TT.includingOptions [ TTO.Option (Proxy @SaveSMT)
...
+getRegName ::
+  Ctx.Index (MS.MacawCrucibleRegTypes MX.X86_64) t ->
+  String
+getRegName reg =
+  case Ctx.intIndex (Ctx.indexVal reg) (Ctx.size MXS.x86RegAssignment) of
+    Just (Some i) ->
+      let r = MXS.x86RegAssignment Ctx.! i
+          rName = MS.crucGenArchRegName MXS.x86_64MacawSymbolicFns r
+      in show rName
+    Nothing -> error "impossible"
```
RyanGlScott, line 84: Consider using `panic` here instead of `error`.

RyanGlScott (approved):
Great stuff! I'm looking forward to being able to use this.

Now that we have `macaw-x86-syntax`, it would also be nice to have `macaw-*-syntax` packages for the other ISAs that `macaw` supports (AArch32, PowerPC, and RISC-V). Could you file an issue about this as a reminder to do this later?

x86_symbolic/tests/Main.hs
langston-barrett: I don't know that it makes a big difference here - The main distinction between `panic` and `error` is the message that gets printed out, especially the confirmation that it is a programmer error (and not, say, a user error). But since this code is only part of the test suite, it's only targeted at Macaw developers, and it's clear from that context that it constitutes a programmer error. In any case, it's real easy to change so I'll go ahead and change it!

langston-barrett:
> Now that we have macaw-x86-syntax, it would also be nice to have macaw-*-syntax packages for the other ISAs that macaw supports (AArch32, PowerPC, and RISC-V). Could you file an issue about this as a reminder to do this later?

Perhaps we just keep #349 open, and add a comment saying that this has been done for x86_64 but the other architectures need it too?

RyanGlScott:
That works for me.

macaw-x86-syntax/README.md
langston-barrett: Ah, this is kind of weird... `X86Regs` is actually something the programmer can write in an S-expression file. `X86Reg` isn't an object-language type, it's just something I'm using as notation to denote the set of syntactic things that can be passed to `get-reg` and `set-reg`... Technically, `{g,s}et-reg` are untypable in Crucible itself, they're kind of dependently-typed in the argument register (consider, e.g., if we had syntax for naming a non-64-bit register).

This weirdness is present in all the `*-syntax` READMEs... Really, they should probably all have Martin-Lof style judgments for the combination of syntactic well-formedness and well-typedness that the parser recognizes, perhaps alongside these more readable, "informal type signatures".

Do you have any thoughts on how we should tackle this? I could at least file an issue upstream in the Crucible repo. The README for `crucible-syntax` is pretty woefully incomplete...

All of this being said, this is almost all just developer-facing, so it's not clear to me how much effort we really want/need to spend precisely specifying these parser's behaviors.

RyanGlScott: Ah, OK. No need to give precise typing rules here: I was just wondering if I had overlooked something. I am fine with being a little hand-wavey here.

x86_symbolic/tests/Main.hs
langston-barrett: Ah, `Data.Macaw.Panic` isn't exposed, and `Lang.Crucible.Panic` feels misleading... I might just leave this as `error` if that's OK with you.

langston-barrett: None of the `Panic` modules in the Macaw repo are exposed, and `Lang.Cruicble.Panic` feels misleading... I might just leave this as `error` if that's alright with you.

RyanGlScott: The usual convention is to create a `PanicComponent` instance for each package so that the `panic` messages can trace back to the specific package that threw the panic (see, for instance, [`Data.Macaw.X86.Symbolic.Panic`](https://github.com/GaloisInc/macaw/blob/1add47389a099b6dedfc6a574a71e2ce433bc617/x86_symbolic/src/Data/Macaw/X86/Symbolic/Panic.hs)).

That being said, I'm fine with `error` as well.

langston-barrett: Is the suggestion that test-suites should have their own `PanicComponent`? Again, it wouldn't be too much effort to add one, but I'm still having a hard time seeing the advantages of adding this over just using `error`. Could you describe the benefits to doing this? I'm happy to implement this (for all the test-suites in the repo) if we can agree it has benefits.

Comparing a few possible approaches:

- We could try to make this function total (e.g., by introducing a `Maybe`). I would argue there's not much point - it's in a test-suite, so doesn't need to be very robust, and it returning `Nothing` would indicate a bug in the test harness (not in an individual test). Thus, I think partiality is both easier and clearer.
- We could have an incomplete pattern match - I think this is basically the same as using `undefined`, and would throw some kind of generic exception?
- We could abort/exit the process - not sure how this would interact with Tasty (e.g., would it be considered a test failure? would other tests run after it fails?)
- We could use `error`, which raises the [`ErrorCall`](https://hackage.haskell.org/package/ghc-internal-9.1001.0/docs/src/GHC.Internal.Exception.html#ErrorCall) exception
- We could define a custom exception type and throw that, like [`Tasty.HUnit` does](https://hackage.haskell.org/package/tasty-hunit-0.10.0.3/docs/src/Test.Tasty.HUnit.Orig.html#assertFailure)
- We could use `panic`, which [itself defines a custom exception type](https://hackage.haskell.org/package/panic-0.4.0.1/docs/Panic.html#t:Panic). It's relatively undocumented, but my impression is that this exception is meant primarily to ask users to file bug reports. This doesn't seem super relevant to me, as users generally don't run the test suite.

All in all, I don't see many advantages nor disadvantages of any of these approaches over one another, and using `error` is really easy :)

RyanGlScott: > Is the suggestion that test-suites should have their own PanicComponent?

Yes. The only real advantage is that you'd get the nice formatting (along with information about where to submit a bug report) that the `panic` library provides. That being said, that's not super important, since we're in a test suite (and most test suite users are developers).

So, again, I'm fine with `error` as well.

langston-barrett: Ok, sorry if I went overboard there :smile: I just wanted to make sure I wasn't missing something important!

Issue #424, opened 2024-08-19
`macaw-base`: `assertPred` does not handle its `isTrue` argument correctly for equalities


Ptival:
https://github.com/GaloisInc/macaw/blob/07db1bad5912404aa32993f5258113451f02b8d8/base/src/Data/Macaw/AbsDomain/JumpBounds.hs#L681-L686

`assertPred` is supposed to enforce that its condition is true when its `isTrue` argument is `True`, and that it is false otherwise.

However, in the `Eq` case, it always enforces that the condition is true. This leads to false rejection of branches which cascades into even more annoying errors.

I will propose a fix soon.

I'm not sure whether it's easy to set a regression for this.

Ptival:
Fixed via #425 ( 9e61b39f985201606a13389e38f0f558a01c5dc0 )

PR #425, opened 2024-08-20
fix `assertPred` and make its code more uniform

Ptival:
This is a fix for #424.  The primary fix consists in splitting the logic for equality tests based on the value of `isTrue`.  I have separated it as its own function as it's needed for two branches, and Haskell lacks binding or-patterns.

Along the way, I have made the code for `assertPred` better documented, and much more uniform.  This should make it easier to spot bugs and understand the overall logic.

### DIFF
--- a/base/src/Data/Macaw/AbsDomain/JumpBounds.hs
+++ b/base/src/Data/Macaw/AbsDomain/JumpBounds.hs
@@ -11,6 +11,7 @@ known equivalent values.
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE RankNTypes #-}
@@ -719,6 +720,48 @@ addWithinRangePred cns w l x u
   | otherwise =
     addRangePred cns (intraStackValueExpr cns x) $! mkRangeBound w (fromInteger l) (fromInteger u)
 
+-- | @addExcludeRangePred cns w x l u@ adds a @(x < l) or (u < x)@ constraint over the
+-- `w`-bit-truncated values.  Fails when the bound excludes all possible values.
+addExcludeRangePred ::
+  RegisterInfo (ArchReg arch) =>
+  BlockIntraStackConstraints arch ids -> NatRepr n -> Value arch ids (BVType n) -> Integer -> Integer ->
+  Either String (BranchConstraints arch ids)
+addExcludeRangePred cns w x l u = do
+  let lowerBoundTooLow = l <= 0              -- (l - 1) would be negative
+  let upperBoundTooHigh = u >= maxUnsigned w -- (u + 1) would overflow
+  if
+    | lowerBoundTooLow && upperBoundTooHigh -> Left "Excluded range excludes all possible values"
+    | lowerBoundTooLow -> addLowerBoundPred cns w (u + 1) x
+    | upperBoundTooHigh -> addUpperBoundPred cns w x (l - 1)
+    | otherwise ->
+      -- compiles it to @(x <= l - 1) or (u + 1 <= x)@
+      disjoinBranchConstraints
+        <$> addUpperBoundPred cns w x (l - 1)
+        <*> addLowerBoundPred cns w (u + 1) x
+
+-- | @assertEqPred cns x w c isTrue@ asserts equality @x = BVValue c w@ is either true or false,
+-- based on `isTrue`.
+assertEqPred ::
+  RegisterInfo (ArchReg arch) =>
+  -- | Constraints that let us evaluate `x`
+  BlockIntraStackConstraints arch ids ->
+  -- | Bitwidth the comparison should apply to
+  NatRepr w ->
+  -- | Value to be constrained
+  Value arch ids (BVType w) ->
+  -- | Numeric constant x should be equal/different to
+  Integer ->
+  -- | `True` if we should assert equality, `False` if we should assert non-equality
+  Bool ->
+  Either String (BranchConstraints arch ids)
+assertEqPred cns w x c isTrue
+  | isTrue =
+    -- x == c becomes c <= x <= c
+    addWithinRangePred cns w c x c
+  | otherwise =
+    -- !(x == c) is handled via !(c <= x <= c) and becomes (x <= c - 1) or (c + 1 <= x)
+    addExcludeRangePred cns w x c c
+
 -- | Assert a predicate is true/false and update bounds.
 --
 -- If it returns a new upper bounds, then that may be used.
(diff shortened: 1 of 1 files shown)


base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -661,6 +661,45 @@ addRangePred cns v p =
         _ ->
           Right $ branchAssignRangePred n p
 
+-- | @assertEqPred cns x w c isTrue@ asserts equality @x = BVValue c w@ is either true, or false,
```
RyanGlScott, line 664:
```suggestion
-- | @assertEqPred cns x w c isTrue@ asserts equality @x = BVValue c w@ is either true or false,
```

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -661,6 +661,45 @@ addRangePred cns v p =
...
+  NatRepr w ->
+  -- | Numeric constant x should be equal/different to
+  Integer ->
+  -- | `True` if we should assert equality, `False` if we should assert non-equality
+  Bool ->
+  Either String (BranchConstraints arch ids)
+assertEqPred cns x w c isTrue
+  | isTrue =
+    addRangePred cns (intraStackValueExpr cns x)
+                     (mkRangeBound w (fromInteger c) (fromInteger c))
```
RyanGlScott, line 686: For symmetry with the `False` case, leave a comment here saying how `x == c` translates to a range bound. I _think_  it's something like `c <= x && x <= c`, but please double-check my understanding.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -661,6 +661,45 @@ addRangePred cns v p =
...
+  | otherwise = do
+    -- !(x == c) becomes:
+    --                   (x >= c + 1)   when c == 0
+    --   (x <= c - 1) || (x >= c + 1)   otherwise
+    -- because ranges use natural numbers and are inclusive on both ends.
```
RyanGlScott, line 691: Do we need a special case for when `c == max`? There is special treatment of this sort in the `EvalApp (BVUnsignedLt (BVValue w c) y)` case in `assertPred`, and I'm not sure if there needs to be similar special treatment here as well.

RyanGlScott (commented):
I'm not terribly familiar with this part of `macaw`, but the general idea makes sense.

It might make sense to split out the `assertPred` refactoring changes into its own commit, followed by a separate commit that does nothing but fix #424.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
Ptival: Yes. I'm going to refactor this even more because I still don't like how it looks.

Ptival: I pushed a solution with a more aggressive refactoring, and split in two commits.

Ptival: Looks like I might have made a mistake. Will investigate...

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -661,89 +661,154 @@ addRangePred cns v p =
         _ ->
           Right $ branchAssignRangePred n p
 
+-- | @addLowerBoundPred cns w l x@ adds a @l <= x@ constraint over the `w`-bit-truncated values.
+-- Fails when the bound excludes all possible values (lower bound greather than the maximum `w`-bit
```
RyanGlScott, line 665:
```suggestion
-- Fails when the bound excludes all possible values (lower bound greater than the maximum `w`-bit
```

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -661,89 +661,154 @@ addRangePred cns v p =
...
+-- value).
+-- Note: The arguments are ordered so that it reads more naturally, as the difference of type
+-- between `l` and `x` prevents mistakes.
+addLowerBoundPred ::
+  RegisterInfo (ArchReg arch) =>
+  BlockIntraStackConstraints arch ids -> NatRepr n -> Integer -> Value arch ids (BVType n) ->
+  Either String (BranchConstraints arch ids)
+addLowerBoundPred cns w l x
+  | l >= maxUnsigned w = Left "Lower bound excludes all possible values"
```
RyanGlScott, line 674: This code confuses me a bit, since the Haddocks claim that it should fail when the lower bound (`l`) is _greater_ than the maximum `w`-bit value (`maxUnsigned w`), but this code (`l >= maxUnsigned w`) will return `True` if `l` is equal to the maximum `w`-bit value. What am I missing?

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -661,89 +661,154 @@ addRangePred cns v p =
...
+-- Fails when the bound excludes all possible values.
+-- Note: The arguments are ordered so that it reads more naturally.
+addWithinRangePred ::
+  RegisterInfo (ArchReg arch) =>
+  BlockIntraStackConstraints arch ids -> NatRepr n -> Integer -> Value arch ids (BVType n) -> Integer ->
+  Either String (BranchConstraints arch ids)
+addWithinRangePred cns w l x u
+  -- Note: we could avoid this duplication by instead producing a conjunction of `addLowerBoundPred`
+  -- and `addUpperBoundPred`, but this produces a single range instead.
+  | l >= maxUnsigned w = Left "Lower bound excludes all possible values"
```
RyanGlScott, line 703: A similar question about `l >= maxUnsigned w` here.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
```diff
@@ -661,89 +661,154 @@ addRangePred cns v p =
...
+  Integer ->
+  -- | `True` if we should assert equality, `False` if we should assert non-equality
+  Bool ->
+  Either String (BranchConstraints arch ids)
+assertEqPred cns w x c isTrue
+  | isTrue =
+    -- x == c becomes c <= x <= c
+    addWithinRangePred cns w c x c
+  | otherwise =
+    -- !(x == c) becomes !(c <= x <= c)
```
RyanGlScott, line 743: Perhaps rewrite `!(c <= x <= c)` as something like `(x < c) or (c < x)`, which is more directly what `addExcludeRangePred` encodes.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
Ptival: Correct, it should be `>`. When `l` is exactly `maxUnsigned w`, this is still a valid range of one possible value. Fixing.

base/src/Data/Macaw/AbsDomain/JumpBounds.hs
Ptival: Found the bug, I needed to be more careful here:
https://github.com/GaloisInc/macaw/blob/45dd9545e5262d201734a9d77df9baec22aaf902/base/src/Data/Macaw/AbsDomain/JumpBounds.hs#L730-L735

RyanGlScott (approved):

PR #426, opened 2024-08-20
ci: Build and test macaw-x86-syntax

langston-barrett:


.github/workflows/ci.yaml
```diff
@@ -107,6 +107,12 @@ jobs:
     - name: Test macaw-x86
       run: cabal test pkg:macaw-x86
 
+    - name: Build macaw-x86-syntax
+      run: cabal build pkg:macaw-x86-syntax
+
+    - name: Test macaw-x86-syntax
+      run: cabal test pkg:macaw-x86-syntax
+
```
RyanGlScott, line 115: I think it would make more sense to run these jobs after the `macaw-x86-symbolic` ones, as `macaw-x86-syntax` depends on `macaw-x86-symbolic`.

.github/workflows/ci.yaml
```diff
@@ -107,6 +107,12 @@ jobs:
...
     - name: Test macaw-x86-symbolic
       if: runner.os == 'Linux'
       run: cabal test pkg:macaw-x86-symbolic
```
RyanGlScott, line 118: Curiously, `macaw-x86-symbolic` is set to only run on Linux, whereas you have `macaw-x86-syntax`'s tests to always run. I'm not really sure why this code is set up this way, as I'm pretty sure that we only run Macaw's CI on Linux machines to begin with. Perhaps we should just remove the `runner.os == 'Linux'` requirements from the CI in a separate commit so that the workflow steps for `macaw-x86-symbolic` and `macaw-x86-syntax` are uniform.

RyanGlScott (approved):

Issue #430, opened 2024-08-27
`symbolic`: Modeling the stack


langston-barrett:
In a downstream project, we're using `macaw-symbolic` to symbolically execute binaries. `macaw-symbolic` encodes functions in the binary as Crucible CFGs that take and return a struct that represents the values of all the registers. To execute such CFGs, we must set initial values for all registers. Most registers can be set to zeroes or symbolic values and plenty of code will run just fine. However, machine code has a lot of expectations about one particular register: the stack pointer.

Each ABI[^1] specifies a stack layout, and our project attempts to model this layout at a low level of abstraction. In particular, we create a large-ish (1MiB) Crucible-LLVM allocation, and:

- On AArch32, the very end is initialized with a number of pointer-sized stack slocks, which are reserved for stack-spilled arguments.[^spill]

- On PPC, the very end is initialized with (in order):

  * A pointer-sized chunk of bytes representing the return address.
  * A pointer-sized zero value, representing the PowerPC back chain for the entrypoint function.
  * A number of pointer-sized stack slocks, which are reserved for stack-spilled arguments.

- On x86_64, the very end is initialized with (in order):

  * A pointer-sized chunk of bytes representing the return address.
  * A number of pointer-sized stack slocks, which are reserved for stack-spilled arguments.

The contents of the stack pointer register are then initialized to be a pointer to this allocation, offset to where the above data begin.

I'd wager that since a lot of code uses the stack, most clients of `macaw-symbolic` want to model the stack *somehow*. Furthermore, since `macaw-symbolic` fairly directly translates manipulations of the stack pointer register in binaries into the Crucible CFGs, I'd argue that this is likely the *only* strategy that works "out of the box". Therefore, I think we should upstream this code (or something like it) into `macaw-*-symbolic`.

One notable *upside* of this kind of model of the stack is that we can examine what happens in the case of certain kinds of common stack-based exploits, such as stack buffer overruns that clobber adjacent variables, as in this C program:
```c
#include "stddef.h"
#include "stdio.h"

__attribute__((noinline)) void zero(char *buf, size_t len) {
  for (int i = 0; i < len; i++) {
    buf[i] = 0;
  }
}

__attribute__((noinline)) void print(char val) {
  printf("val = %i\n", val);
}

void caller(void) {
  char val = 1;
  char buf[8] = {1};
  zero(buf, 9);
  print(val);
}

int main() {
  caller();
}
```
[...]

PR #431, opened 2024-08-29
symbolic: Factor stack setup code into its own module

langston-barrett:
Factors some code out of `Testing` into its own module for the sake of clarity and code-sharing. Several downstream projects could benefit from this API, see #430 for examples.

Also, rewrites a comment that I believe to have been misleading.

### DIFF
--- a/symbolic/src/Data/Macaw/Symbolic/Stack.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/Stack.hs
@@ -15,11 +15,8 @@ module Data.Macaw.Symbolic.Stack
   ) where
 
 import Data.BitVector.Sized qualified as BVS
-import Data.Macaw.CFG qualified as MC
-import Data.Macaw.Symbolic qualified as MS
 import Data.Parameterized.Context as Ctx
 import Lang.Crucible.Backend qualified as C
-import Lang.Crucible.Simulator qualified as C
 import Lang.Crucible.LLVM.DataLayout qualified as CLD
 import Lang.Crucible.LLVM.MemModel qualified as CLM
 import What4.Interface qualified as WI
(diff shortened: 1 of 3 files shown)


symbolic/src/Data/Macaw/Symbolic/Stack.hs
```diff
@@ -0,0 +1,115 @@
...
+import What4.Interface qualified as WI
+import What4.Symbol qualified as WSym
+
+-- | Helper, not exported
+stackAlign :: CLD.Alignment
+stackAlign = CLD.noAlignment
+
+-- | Create an SMT array representing the program stack.
+--
+-- Builds a 'WI.freshConstant' with name "stack_array".
```
RyanGlScott, line 31: These are rather low-level implementation details. I think that for the sake of Haddocks, it would suffice to say that you're creating a fresh SMT array.

symbolic/src/Data/Macaw/Symbolic/Stack.hs
```diff
@@ -0,0 +1,115 @@
...
+
+-- | Allocate this many pointer-sized stack slots, which are reserved for
+-- stack-spilled arguments.
+newtype ExtraStackSlots = ExtraStackSlots { getExtraStackSlots :: Int }
+  -- Derive the Read instance using the `newtype` strategy, not the
+  -- `stock` strategy. This allows parsing ExtraStackSlots values in
+  -- optparse-applicative-based command-line parsers using the `Read` instance.
+  deriving newtype (Enum, Eq, Integral, Num, Ord, Read, Real, Show)
+
+-- | An allocartion representing the program stack, backed by an SMT array
```
RyanGlScott, line 72:
```suggestion
-- | An allocation representing the program stack, backed by an SMT array
```

symbolic/src/Data/Macaw/Symbolic/Stack.hs
```diff
@@ -0,0 +1,115 @@
...
+data ArrayStack sym w
+  = ArrayStack
+    { -- | Pointer to the base of the stack array
+      arrayStackBasePtr :: CLM.LLVMPtr sym w
+      -- | Pointer to the top (end) of the stack array
+    , arrayStackTopPtr :: CLM.LLVMPtr sym w
+      -- | SMT array backing the stack allocation
+    , arrayStackVal :: WI.SymArray sym (Ctx.SingleCtx (WI.BaseBVType w)) (WI.BaseBVType 8)
```
RyanGlScott, line 80: Do you foresee it being convenient to have different underlying stack representations besides an SMT array? (Perhaps that's not in scope for this PR, but I thought I'd bring it up given your comments in https://github.com/GaloisInc/macaw/issues/430#issuecomment-2318712835.)

RyanGlScott (approved):

symbolic/src/Data/Macaw/Symbolic/Stack.hs
langston-barrett: I think it *might* be the case that the SMT array memory model can't detect uninitialized reads? I might be misremembering that, though. Frankly, I don't understand the trade-offs between the models, so I can't really comment on whether or not it would be useful to have both representations. In either case, doesn't hurt to be explicit here.

PR #433, opened 2024-09-03
 x86-symbolic: Setting up a SysV-compatible stack 

langston-barrett:
Towards #430 and #434.

### DIFF
--- a/macaw-aarch32-symbolic/tests/README.org
+++ b/macaw-aarch32-symbolic/tests/README.org
@@ -1,7 +1,7 @@
 * Overview
 This test suite tests that symbolic execution of AArch32 programs works. It is also a convenient place to test the semantics.  The test harness looks for binaries in the ~pass~ and ~fail~ subdirectories.  In both cases, it enumerates the functions whose names are prefixed by ~test_~, each of which are symbolically executed with fully symbolic initial states.  It treats the return value of each test function as an assertion that it will attempt to prove (i.e., it checks the satisfiability of the assertion).  Test cases in ~pass~ are expected to return a value that is always true, while the test cases in ~fail~ are expected to not always be true (but note that the precise models are not inspected).  The test harness uses yices for path satisfiability checking and z3 to prove goals.  Since path satisfiability checking is enabled, test cases can include loops as long as they symbolically terminate.
 
-In addition to proving that the result of a test case is always true (or not), the test harness will attempt to prove all generated side conditions for tests with names prefixed by ~test_and_verify_~. In both the ~pass~ and ~fail~ directories, these side conditions are expected to pass (or they will cause test failures).
+In addition to proving that the result of a test case is always true (or not), the test harness will attempt to prove all generated side conditions. In both the ~pass~ and ~fail~ directories, these side conditions are expected to pass (or they will cause test failures).
 
 * Usage
 
--- a/symbolic/src/Data/Macaw/Symbolic/Testing.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/Testing.hs
@@ -475,6 +475,11 @@ simulateAndVerify goalSolver logger bak execFeatures archInfo archVals binfo mmP
 
     summarizeExecution goalSolver logger bak memVar extractor execResult
 
+-- | Post-process the results of symbolic execution
+--
+-- In particular, turn a 'CS.ExecResult' into a 'SimulationResult'. If
+-- simulation succeeded, uses the 'ResultExtractor' to assert that the executed
+-- function returned a nonzero value.
 summarizeExecution ::
   ( CB.IsSymBackend sym bak
   , CCE.IsSyntaxExtension (MS.MacawExt arch)
@@ -587,9 +592,14 @@ lazyInitialMem binfo bak archInfo archVals = do
                  }
   pure (InitialMem mem mmConf)
 
+-- | Initial program memory and memory configuration
+--
+-- Returned by 'initialMem' and 'lazyInitialMem'.
 data InitialMem p sym arch
   = InitialMem
-    { initMemMem :: CLM.MemImpl sym
+    { -- | Initial contents of memory, before symbolic execution
+      initMemMem :: CLM.MemImpl sym
+      -- | Memory model configuration
     , initMemMmConf :: MS.MemModelConfig p sym arch CLM.Mem
     }
 
@@ -628,7 +638,7 @@ simulateFunction :: forall arch sym bak w ext ids
                        )
 simulateFunction bak execFeatures archVals halloc iMem binfo dfi = do
   let ?ptrWidth = WI.knownRepr
-  InitialMem initMem mmConf <- pure iMem
+  let InitialMem initMem mmConf = iMem
   (regs, mem) <- defaultRegs bak archVals initMem
   let iMem' = InitialMem mem mmConf
   simDiscoveredFunction bak execFeatures archVals halloc iMem' regs binfo dfi
@@ -656,7 +666,7 @@ simDiscoveredFunction ::
      )
 simDiscoveredFunction bak execFeatures archVals halloc iMem regs binfo dfi = do
   let sym = CB.backendGetSym bak
-  InitialMem mem mmConf <- pure iMem
+  let InitialMem mem mmConf = iMem
   memVar <- CLM.mkMemVar "macaw-symbolic:test-harness:llvm_memory" halloc
   extImpl <-
     MS.withArchEval archVals sym $ \archEvalFn ->
--- a/symbolic/src/Data/Macaw/Symbolic/Testing.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/Testing.hs
@@ -42,7 +42,6 @@ module Data.Macaw.Symbolic.Testing (
 import qualified Control.Exception as X
 import qualified Control.Lens as L
 import           Control.Lens ( (&), (%~) )
-import           Control.Monad ( when )
 import           Control.Monad.Except ( runExceptT )
 import qualified Data.Bits as Bits
 import qualified Data.BitVector.Sized as BVS
@@ -415,11 +414,7 @@ proveGoals bak goalSolver = do
 -- architecture-specific primitives).
 --
 -- In addition to symbolically executing the function to produce a Sat/Unsat
--- result, this function will attempt to verify all generate side conditions if
--- the name of the function being simulated begins with @test_and_verify_@ (as
--- opposed to just @test@).  This is necessary for testing aspects of the
--- semantics that involve the generated side conditions, rather than just the
--- final result.
+-- result, this function will attempt to verify all generated side conditions.
 simulateAndVerify :: forall arch sym bak ids w solver scope st fs
                    . ( CB.IsSymBackend sym bak
                      , CCE.IsSyntaxExtension (MS.MacawExt arch)
@@ -477,7 +472,7 @@ simulateAndVerify goalSolver logger bak execFeatures archInfo archVals binfo mmP
     (memVar, execResult) <-
       simulateFunction bak execFeatures archVals halloc iMem g
 
-    summarizeExecution funName goalSolver logger bak memVar extractor execResult
+    summarizeExecution goalSolver logger bak memVar extractor execResult
 
 summarizeExecution :: 
   ( CB.IsSymBackend sym bak
@@ -490,7 +485,6 @@ summarizeExecution ::
   , bak ~ CBO.OnlineBackend solver scope st fs
   , WPO.OnlineSolver solver
   ) =>
-  WF.FunctionName ->
   -- | The solver adapter to use to discharge assertions
   WS.SolverAdapter st ->
   -- | A logger to (optionally) record solver interaction (for the goal solver)
@@ -501,7 +495,7 @@ summarizeExecution ::
   ResultExtractor sym arch ->
   CS.ExecResult (MS.MacawLazySimulatorState sym w) sym ext (CS.RegEntry sym (MS.ArchRegStruct arch)) ->
   IO SimulationResult
-summarizeExecution funName goalSolver logger bak memVar (ResultExtractor withResult) =
+summarizeExecution goalSolver logger bak memVar (ResultExtractor withResult) =
   \case
     CS.TimeoutResult {} -> return SimulationTimeout
     CS.AbortedResult {} -> return SimulationAborted
@@ -509,8 +503,7 @@ summarizeExecution funName goalSolver logger bak memVar (ResultExtractor withRes
       case partialRes of
         CS.PartialRes {} -> return SimulationPartial
         CS.TotalRes gp -> do
-          when ("test_and_verify_" `Text.isPrefixOf` WF.functionName funName) $
-            proveGoals bak goalSolver
+          proveGoals bak goalSolver
 
           postMem <- case CSG.lookupGlobal memVar (gp L.^. CS.gpGlobals) of
                        Just postMem -> pure postMem
--- a/symbolic/src/Data/Macaw/Symbolic/Testing.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/Testing.hs
@@ -314,6 +314,55 @@ freshRegs symArchFns sym =
     (mkInitialRegVal symArchFns sym)
     (MS.crucGenRegAssignment symArchFns)
 
+-- | Make initial registers, including setting up a stack pointer.
+--
+-- The stack pointer points to the middle of the stack allocation. This is a
+-- hack. We do this because each architecture has some expected stack layout
+-- (e.g., x86_64 expects a return address just above the stack pointer;
+-- PPC expects a "back chain"; and AArch32, PPC, and x86_64 all expect
+-- stack-spilled arguments above everything else). Setting the stack pointer
+-- to the middle of the allocation allows reads of fully symbolic data from
+-- above the stack pointer, and this seems to be good enough to make our tests
+-- pass.
+--
+-- The functions in the test-suite do not (and should not) rely on accessing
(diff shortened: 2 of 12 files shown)


x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
```diff
@@ -472,6 +480,102 @@ instance GenArchInfo LLVMMemory M.X86_64 where
...
+
+-- | Create an allocation for the stack for x86_64 with the SysV ABI.
+--
+-- On x86_64 with the SysV ABI, the stack grows \"downwards\" from high
+-- addresses to low. The end of the stack is initialized with the ELF auxiliary
+-- vector (not modelled here), and functions expect the following data to be
+-- available above their stack frame (i.e., just above the address in @rsp@):
+--
+-- * Their stack-spilled arguments
+-- * The return address
```
RyanGlScott, line 514: Perhaps these should be listed in the opposite order, since the return address is in `0(%rsp)`, followed by the stack-spilled arguments in `8(%rsp)`, `16(%rsp)`, and so on.

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
```diff
@@ -472,6 +480,102 @@ instance GenArchInfo LLVMMemory M.X86_64 where
...
+x86SysvStack ::
+  C.IsSymBackend sym bak =>
+  (?memOpts :: MM.MemOptions) =>
+  MM.HasLLVMAnn sym =>
+  bak ->
+  MM.MemImpl sym ->
+  -- | Initialize the end of the stack to a concrete sequence of bytes if the
+  -- value is @Just@, otherwise initialize the end of the stack to 8 fresh,
+  -- symbolic bytes.
+  Maybe (Seq.Seq Word8) ->
```
RyanGlScott, line 544: I believe it is always the case that the return address is a 64-bit word. If so, it might be better to model this as a `BV 64` instead of as a `Seq Word8` to avoid unrepresentable states.

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
```diff
@@ -472,6 +480,102 @@ instance GenArchInfo LLVMMemory M.X86_64 where
...
+  MSS.ExtraStackSlots ->
+  -- | Size of stack allocation
+  WI.SymExpr sym (WI.BaseBVType 64) ->
+  IO (MSS.ArrayStack sym 64, MM.MemImpl sym)
+x86SysvStack bak mem topBytes slots sz = do
+  let ptrBytes = 8
+  let ?ptrWidth = C.knownNat @64
+
+  (arrayStack, mem') <- MSS.createArrayStack bak mem slots sz
+  MSS.ArrayStack _base top _array <- pure arrayStack
```
RyanGlScott, line 555: A bit simpler:

```suggestion
  (arrayStack@(MSS.ArrayStack _base top _array), mem') <- MSS.createArrayStack bak mem slots sz
```

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
```diff
@@ -472,6 +480,102 @@ instance GenArchInfo LLVMMemory M.X86_64 where
...
+  -- | The caller must ensure the size of these is less than the allocation size
+  MSS.ExtraStackSlots ->
+  -- | Size of stack allocation
+  WI.SymExpr sym (WI.BaseBVType 64) ->
+  IO (MSS.ArrayStack sym 64, MM.MemImpl sym)
+x86SysvStack bak mem topBytes slots sz = do
+  let ptrBytes = 8
+  let ?ptrWidth = C.knownNat @64
+
+  (arrayStack, mem') <- MSS.createArrayStack bak mem slots sz
```
RyanGlScott, line 554: I think there should be a `-- (1)` comment above this line for consistency with the `-- (2)` and `-- (3)` comments below.

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
```diff
@@ -472,6 +480,102 @@ instance GenArchInfo LLVMMemory M.X86_64 where
...
+  (?memOpts :: MM.MemOptions) =>
+  MM.HasLLVMAnn sym =>
+  bak ->
+  MM.MemImpl sym ->
+  MM.LLVMPtr sym w ->
+  Seq.Seq Word8 ->
+  IO (MM.LLVMPtr sym w, MM.MemImpl sym)
+writeBytesBackwards bak mem ptr bytes = do
+  let sym = C.backendGetSym bak
+  let i8 = MM.bitvectorType (Bytes.toBytes (1 :: Int))
```
RyanGlScott, line 496: It might be worth factoring out `i8 :: StorageType` into its own helper definition, as it's used both here and in `x86SysvStack`.

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
```diff
@@ -472,6 +480,102 @@ instance GenArchInfo LLVMMemory M.X86_64 where
     , updateReg = x86UpdateReg
     }
 
+-- | Helper, not exported
+writeBytesBackwards ::
```
RyanGlScott, line 484: I think it would be helpful to state in the comments that:

1. This writes each byte in `bytes` at successively smaller negative offsets from `ptr`.
2. This assumes the precondition that `ptr`'s allocation is large enough to store the data in `bytes`.

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
langston-barrett: Yep, I'm refactoring all this, and plan to do this.

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
langston-barrett: I'm going to change all the code in this PR significantly, this will not be as relevant then.

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
langston-barrett: Will make the relation of order to addresses more explicit in Haddocks.

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
langston-barrett: No longer relevant.

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
langston-barrett: No longer relevant.

x86_symbolic/src/Data/Macaw/X86/Symbolic.hs
langston-barrett: I've removed this function, will add haddocks to functions that make assertions

langston-barrett:
Ok, I finally got this to the point where I can run the SysV stack setup code in the `macaw-x86-symbolic` test suite. Tests are now failing mysteriously:
```
        tests/fail/constant.unopt.exe:      FAIL
          Testing test_constant
            You have encountered a bug in Crucible's implementation.
            *** Please create an issue at https://github.com/GaloisInc/crucible/issues

            %< ---------------------------------------------------
              Revision:  e86140641260b00f64fdbd51d73d48d766c55bdf
              Branch:    HEAD
              Location:  writeMemWithAllocationCheck
              Message:   Expected succesful byte load when updating SMT array
                         but got an error instead
              panic, called at src/Lang/Crucible/Panic.hs:11:9 in crucible-0.7.0.99-inplace:Lang.Crucible.Panic
              panic, called at src/Lang/Crucible/LLVM/MemModel/Generic.hs:1333:23 in crucible-llvm-0.6.0.99-inplace:Lang.Crucible.LLVM.MemModel.Generic
            %< ---------------------------------------------------

          Use -p '/Lazy memory model.False assertions.tests\/fail\/constant.unopt.exe/' to rerun this test only.
```
https://github.com/GaloisInc/crucible/blob/62359553c204abd18c4d98fcff74b750a2a15232/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs#L1333-L1336

langston-barrett:
Ok, it's actually kind of non-trivial to see *why* this is failing. I modified `?recordLLVMAnnotation` to print out the memory error explanations as the safety assertions are made, the last one for each test-case is something like this, so I *think* this is the one that's failing:
```
Unexpected argument type:
While selecting the low bits of a bitvector
  bitvectorType 8
Performing store at type: i512
  Via pointer: (2, 0xffff8:[64])
In memory state:
  Base memory
    Allocations:
      StackAlloc 2 0x100000:[64] Mutable 1-byte-aligned stack_alloc
      GlobalAlloc 1 0xffffffffffffffff:[64] Mutable 1-byte-aligned Global memory for macaw-symbolic
    Writes:
      Indexed chunk:
        1 |->   *(1, 0x0:[64]) := cglobalMemoryBytes@0:a
        2 |->   *(2, 0x0:[64]) := cstack_array@76:a
```
It makes sense that this would fail, because it's performing a 512-bit store at a stack offset that definitely doesn't have enough room for that. As for *why* it's doing a 512-bit store there... :shrug:

langston-barrett:
> As for why it's doing a 512-bit store there... 🤷

It's because I wrote `Bytes.toBytes 64` when I meant `Bytes.toBytes 8` when describing the size of the return address that gets written to the stack. Fixed! Still seems wrong for the memory model to panic, I would expect for an assertion to fail with a reasonable message.

macaw-aarch32-symbolic/tests/README.org
```diff
@@ -1,7 +1,7 @@
...
 
-In addition to proving that the result of a test case is always true (or not), the test harness will attempt to prove all generated side conditions for tests with names prefixed by ~test_and_verify_~. In both the ~pass~ and ~fail~ directories, these side conditions are expected to pass (or they will cause test failures).
+In addition to proving that the result of a test case is always true (or not), the test harness will attempt to prove all generated side conditions. In both the ~pass~ and ~fail~ directories, these side conditions are expected to pass (or they will cause test failures).
```
RyanGlScott, line 4: What is the motivation for this change? (I think I'm fine with making each test case prove all side conditions, but I'm wondering if there was a specific reason for doing so as part of this work.)

symbolic/src/Data/Macaw/Symbolic/Testing.hs
```diff
@@ -401,46 +456,142 @@ simulateAndVerify :: forall arch sym bak ids w solver scope st fs
...
+      case mmPreset of
+        DefaultMemModel -> initialMem binfo bak archInfo archVals
+        LazyMemModel -> lazyInitialMem binfo bak archInfo archVals
+
+    (memVar, execResult) <-
+      simulateFunction bak execFeatures archVals halloc iMem binfo dfi
+
+    summarizeExecution goalSolver logger bak memVar extractor execResult
+
+summarizeExecution ::
```
RyanGlScott, line 478: Can you add some Haddocks to this function? It's exported, so we might as well.

symbolic/src/Data/Macaw/Symbolic/Testing.hs
```diff
@@ -401,46 +456,142 @@ simulateAndVerify :: forall arch sym bak ids w solver scope st fs
...
+    MSMLazy.newMergedGlobalMemoryWith populateRelocation archInfo bak
+      endianness MSMLazy.ConcreteMutable (binariesMems binfo)
+  let mmConf = (MSMLazy.memModelConfig bak memPtrTbl)
+                 { MS.lookupFunctionHandle = lookupFunction archVals binfo
+                 , MS.lookupSyscallHandle = lookupSyscall
+                 , MS.mkGlobalPointerValidityAssertion = validityCheck
+                 }
+  pure (InitialMem mem mmConf)
+
+data InitialMem p sym arch
```
RyanGlScott, line 590: Similarly, Haddocks plz

symbolic/src/Data/Macaw/Symbolic/Testing.hs
```diff
@@ -449,121 +600,100 @@ simulateAndVerify goalSolver logger bak execFeatures archInfo archVals binfo mmP
...
-  -- The functions in the test-suite do not (and should not) rely on accessing
-  -- data in their caller's stack frames, even though that wouldn't cause test
-  -- failures with this setup.
-  initialRegs <- MS.macawAssignToCrucM (mkInitialRegVal symArchFns sym) (MS.crucGenRegAssignment symArchFns)
-  stackInitialOffset <- WI.bvLit sym WI.knownRepr (BVS.mkBV WI.knownRepr mib)
-  sp <- CLM.ptrAdd sym WI.knownRepr stackBasePtr stackInitialOffset
-  let initialRegsEntry = CS.RegEntry regsRepr initialRegs
-  let regsWithStack = MS.updateReg archVals initialRegsEntry MC.sp_reg sp
-
+  InitialMem initMem mmConf <- pure iMem
```
RyanGlScott, line 631: Using `pure` here just to pattern match on `InitialMem` feels silly. What not just:

```suggestion
  let InitialMem initMem mmConf = iMem
```

symbolic/src/Data/Macaw/Symbolic/Testing.hs
```diff
@@ -449,121 +600,100 @@ simulateAndVerify goalSolver logger bak execFeatures archInfo archVals binfo mmP
...
+  CS.RegEntry sym (MS.ArchRegStruct arch) ->
+  BinariesInfo arch ->
+  MD.DiscoveryFunInfo arch ids ->
+  -- ^ The function to simulate
+  IO ( CS.GlobalVar CLM.Mem
+     , CS.ExecResult (MS.MacawLazySimulatorState sym w) sym ext (CS.RegEntry sym (MS.ArchRegStruct arch))
+     )
+simDiscoveredFunction bak execFeatures archVals halloc iMem regs binfo dfi = do
+  let sym = CB.backendGetSym bak
+  InitialMem mem mmConf <- pure iMem
```
RyanGlScott, line 659: Similarly:

```suggestion
  let InitialMem mem mmConf = iMem
```

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
```diff
@@ -0,0 +1,219 @@
...
+constraints. ABI-compatible machine code translated by Macaw manages the stack
+for itself, so this module is primarily helpful for initial setup of the stack,
+before starting symbolic execution.
+
+Helpful links:
+
+* <https://eli.thegreenplace.net/2011/09/06/stack-frame-layout-on-x86-64>
+* <https://wiki.osdev.org/System_V_ABI#x86-64>
+
+TODO: The stack is supposed to be 16-byte aligned before a @call@.
```
RyanGlScott, line 24: Does this TODO still need to be addressed?

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
```diff
@@ -0,0 +1,219 @@
...
+  Lens.Lens'
+    (Ctx.Assignment (C.RegValue' sym) (MS.MacawCrucibleRegTypes X86.X86_64))
+    (StackPointer sym)
+stackPointerReg =
+  Lens.lens
+    (\regs -> StackPointer (C.unRV (regs Lens.^. ixF' X86S.rsp)))
+    (\regs v -> regs Lens.& ixF' X86S.rsp Lens..~ C.RV (getStackPointer v))
+
+-- | A return address
+newtype RetAddr sym = RetAddr { getRetAddr :: MM.LLVMPtr sym 64 }
```
RyanGlScott, line 85: Is there ever a scenario where you'd want the block number for the return address's `LLVMPtr` to be non-zero? If not, wouldn't it be more direct to represent this as a `SymBV sym 64` instead?

macaw-aarch32-symbolic/tests/README.org
RyanGlScott: Gotcha.

Part of me wonders if we should rename the `test_and_verify_*` functions in the test cases to simply `test_*` now that we are no longer using the former convention. Then again, I don't think it's recompiling the binaries (which would, in turn, bloat the `git` history) just for this reason, so perhaps better to leave well enough alone.

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
langston-barrett: I can't think of one right now, but perhaps in the future we would support something like `mmap`ing new code allocations, e.g., in the context of JIT compilation, and would support `ret`ing to such allocations? I'm not sure there's a super clear advantage to disallowing it, other than making the more common usage (block number 0) more obvious.

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
langston-barrett: Yes, but I might prefer to do so in a follow-up if that's alright. I'm not exactly sure what this even looks like (e.g., do you always put the return address 8 bytes before a 16-byte-aligned address?), so I'd have to do a bit more digging to figure it out. In any case, it doesn't seem to affect the test suite.

RyanGlScott (approved):

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
langston-barrett: https://github.com/GaloisInc/macaw/issues/436

PR #437, opened 2024-09-19
x86-symbolic: Establish SysV stack alignment

### LINKED ISSUE #436: x86-symbolic: Stack alignment
On the subject of the stack under the x86_64 SysV ABI, the [OSDev wiki says](https://wiki.osdev.org/System_V_ABI):

> The stack is 16-byte aligned just before the call instruction is called.

We should ensure that the functions introduced in #433 properly establish this precondition.

langston-barrett:
Fixes #436.

### DIFF
--- a/x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
+++ b/x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
@@ -30,7 +30,7 @@ High addresses
 |---------------------|
 | ...                 |
 |---------------------|
-| Spilled argument 1  |
+| Spilled argument 2  |
 |---------------------|
 | Spilled argument 1  |
 |---------------------| <- %rsp + 8 (16-byte aligned)
@@ -112,6 +112,16 @@ stackPointerReg =
   Lens.lens
     (\regs -> StackPointer (C.unRV (regs Lens.^. ixF' X86S.rsp)))
     (\regs v -> regs Lens.& ixF' X86S.rsp Lens..~ C.RV (getStackPointer v))
+
+-- | Allocate stack space by subtracting from the stack pointer
+allocStackSpace ::
+  C.IsSymInterface sym =>
+  sym ->
+  StackPointer sym ->
+  WI.SymBV sym 64 ->
+  IO (StackPointer sym)
+allocStackSpace sym (StackPointer sp) amount =
+  StackPointer <$> MM.ptrSub sym ptrRepr sp amount
 
 -- | A return address
 newtype RetAddr sym = RetAddr { getRetAddr :: MM.LLVMPtr sym 64 }
@@ -197,16 +207,16 @@ writeSpilledArgs bak mem sp spilledArgs = do
   -- +----------------------+-------------+
   --
   let align16 = CLD.exponentToAlignment 4  -- 2^4 = 16
-  StackPointer spAligned16 <- alignStackPointer sym align16 sp
+  sp16@(StackPointer spAligned16) <- alignStackPointer sym align16 sp
   isAligned16 <- MM.ptrEq sym ptrRepr spAligned8 spAligned16
   sp' <-
     if even (Seq.length (getSpilledArgs spilledArgs))
     then
       -- In this case, either the stack pointer is *already* 16-byte aligned, or
       -- it *needs to be* 16-byte aligned. In either case, this value suffices.
-      pure (StackPointer spAligned16)
+      pure sp16
     else do
-      spAdd8 <- MM.ptrAdd sym ptrRepr spAligned16 eight
+      StackPointer spAdd8 <- allocStackSpace sym sp16 eight
       StackPointer <$> MM.muxLLVMPtr sym isAligned16 spAdd8 spAligned8
 
   let writeWord (StackPointer p, m) arg = do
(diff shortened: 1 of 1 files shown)


x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
```diff
@@ -8,9 +8,40 @@ modelled here), and functions expect the following data to be available above
...
+| Padding (if needed) |
+|---------------------|
+| Spilled argument n  |
+|---------------------|
+| ...                 |
+|---------------------|
+| Spilled argument 1  |
+|---------------------|
+| Spilled argument 1  |
+|---------------------| <- %rsp + 8 (16-byte aligned)
```
RyanGlScott, line 36:
```suggestion
|---------------------|
| Spilled argument 2  |
|---------------------|
| Spilled argument 1  |
|---------------------| <- %rsp + 8 (16-byte aligned)
```

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
```diff
@@ -91,14 +123,36 @@ freshRetAddr sym = do
...
 -- | Stack-spilled arguments, in normal order
 newtype SpilledArgs sym
   = SpilledArgs { getSpilledArgs :: Seq.Seq (MM.LLVMPtr sym 64) }
 
 -- | Write pointer-sized stack-spilled arguments to the stack.
 --
 -- SysV specifies that they will be written in reverse order, i.e., the last
--- element of the 'Seq.Seq' will be written to the highest address.
+-- element of the 'Seq.Seq' will be written to the highest address. It further
+-- specifies that tne end of the argument list will be 16-byte aligned. This
```
RyanGlScott, line 153:
```suggestion
-- specifies that the end of the argument list will be 16-byte aligned. This
```

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
```diff
@@ -116,11 +170,50 @@ writeSpilledArgs bak mem sp spilledArgs = do
...
+
+  -- First, align to 8 bytes. In all probability, this is a no-op.
+  let align8 = CLD.exponentToAlignment 3  -- 2^3 = 8
+  StackPointer spAligned8 <- alignStackPointer sym align8 sp
+
+  -- At this point, exactly one of `spAligned8` or `spAligned8 + 8` is 16-byte
+  -- aligned. We need to ensure that, after writing the argument list, the stack
+  -- pointer will be 16-byte aligned.
+  --
+  -- Depending on the current alignment and partity of the number of spilled
```
RyanGlScott, line 182:
```suggestion
  -- Depending on the current alignment and parity of the number of spilled
```

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
```diff
@@ -116,11 +170,50 @@ writeSpilledArgs bak mem sp spilledArgs = do
...
+  -- arguments, we need to:
+  --
+  -- - further align `rsp` to 16 bytes (call this choice a16)
+  -- - add 8 so that `rsp + 8` is aligned to 16 bytes (call this +8)
```
RyanGlScott, line 186: Given that the stack grows downwards, I'm somewhat surprised that you'd _add_ 8 to align the stack pointer to 16 bytes. Shouldn't you _subtract_ 8 instead? (I might be missing something here.)

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
```diff
@@ -116,11 +170,50 @@ writeSpilledArgs bak mem sp spilledArgs = do
...
+  StackPointer spAligned8 <- alignStackPointer sym align8 sp
+
+  -- At this point, exactly one of `spAligned8` or `spAligned8 + 8` is 16-byte
+  -- aligned. We need to ensure that, after writing the argument list, the stack
+  -- pointer will be 16-byte aligned.
+  --
+  -- Depending on the current alignment and partity of the number of spilled
+  -- arguments, we need to:
+  --
+  -- - further align `rsp` to 16 bytes (call this choice a16)
```
RyanGlScott, line 185: Given that the stack pointer is guaranteed to be 8-byte-aligned at this point, is there any meaningful difference between "futher align `rsp` to 16 bytes" and "subtract 8 from `rsp`"? I'm wondering if we could combine the `a16` and `+8` cases.

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
langston-barrett: Whoops, yep! Added a new function for allocating stack space that makes this more explicit.

x86_symbolic/src/Data/Macaw/X86/Symbolic/ABI/SysV.hs
langston-barrett: No, I agree that there is no meaningful difference. I've updated the comments to reflect this, though I don't know if this insight can help us actually simplify the implementation.

RyanGlScott (approved):

PR #438, opened 2024-09-20
aarch32-symbolic: Named `Index`es into `ArchRegContext AArch32`

langston-barrett:
... like the ones for x86_64. These will help in development of a concrete syntax for macaw-symbolic-aarch32, and can be helpful when setting up or interpreting the results from symbolic execution.

PR #439, opened 2024-09-20
aarch32-symbolic: Setting up an ABI-compatible stack

langston-barrett:
Like #433 + #437, but for AArch32. Towards #430. Based on #438. Needs to be integrated into the test suite for validation.

### DIFF
--- a/macaw-aarch32-symbolic/src/Data/Macaw/AArch32/Symbolic/ABI.hs
+++ b/macaw-aarch32-symbolic/src/Data/Macaw/AArch32/Symbolic/ABI.hs
@@ -68,10 +68,10 @@ module Data.Macaw.AArch32.Symbolic.ABI
   ) where
 
 import Control.Lens qualified as Lens
-import Control.Monad qualified as Monad
 import Data.BitVector.Sized qualified as BVS
-import Data.Macaw.ARM qualified as AArch32
+import Data.Coerce (coerce)
 import Data.Macaw.AArch32.Symbolic qualified as AArch32S
+import Data.Macaw.ARM qualified as AArch32
 import Data.Macaw.Symbolic qualified as MS
 import Data.Macaw.Symbolic.Stack qualified as MSS
 import Data.Parameterized.Classes (ixF')
@@ -79,7 +79,6 @@ import Data.Parameterized.Context qualified as Ctx
 import Data.Sequence qualified as Seq
 import GHC.Natural (naturalToInteger)
 import Lang.Crucible.Backend qualified as C
-import Lang.Crucible.LLVM.Bytes qualified as Bytes
 import Lang.Crucible.LLVM.DataLayout qualified as CLD
 import Lang.Crucible.LLVM.MemModel qualified as MM
 import Lang.Crucible.Simulator qualified as C
@@ -101,16 +100,6 @@ stackPointerReg =
   Lens.lens
     (\regs -> StackPointer (C.unRV (regs Lens.^. ixF' AArch32S.sp)))
     (\regs v -> regs Lens.& ixF' AArch32S.sp Lens..~ C.RV (getStackPointer v))
-
--- | Allocate stack space by subtracting from the stack pointer
-allocStackSpace ::
-  C.IsSymInterface sym =>
-  sym ->
-  StackPointer sym ->
-  WI.SymBV sym 32 ->
-  IO (StackPointer sym)
-allocStackSpace sym (StackPointer sp) amount =
-  StackPointer <$> MM.ptrSub sym ptrRepr sp amount
 
 -- | Align the stack pointer to a particular 'CLD.Alignment'.
 alignStackPointer ::
@@ -137,9 +126,9 @@ newtype SpilledArgs sym
 
 -- | Write pointer-sized stack-spilled arguments to the stack.
 --
--- SysV specifies that they will be written in reverse order, i.e., the last
+-- The ABI specifies that they will be written in reverse order, i.e., the last
 -- element of the 'Seq.Seq' will be written to the highest address. It further
--- specifies that the end of the argument list will be 16-byte aligned. This
+-- specifies that the end of the argument list will be 8-byte aligned. This
 -- function will allocate additional stack space before the spilled arguments if
 -- necessary to establish this constraint.
 --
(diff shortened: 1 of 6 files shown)


macaw-aarch32-symbolic/src/Data/Macaw/AArch32/Symbolic/ABI.hs
```diff
@@ -0,0 +1,249 @@
+{-|
```
langston-barrett, line 1: @RyanGlScott There's a lot of code here that's basically equivalent to the code for the x86_64 SysV ABI, particularly the part about writing spilled arguments to the stack while maintaining proper alignment. Do you have thoughts on the best way to share between the two (or if we should share it)? I could lift `writeSpilledArgs` to `Data.Macaw.Symbolic.Stack` perhaps, with a note that it's only appropriate for use on certain ABIs, and that you'd probably want to use the specializations of it that we would provide in `Data.Macaw.{X86,AArch32}.Symbolic.ABI`?

RyanGlScott: I think putting the shared code (with caveats that it is only valid for certain ABIs) in `macaw-symbolic` is fine.

langston-barrett:
Here's the source and the disassembly:
```c
// movt.c
// A test case which ensures that the `movt` instruction works as expected.
int __attribute__((noinline)) test_movt(void) {
  int ret = 0;
  // After the movt instruction, the value of r0 should be 0x10000, which
  // should suffice to make the `movne %0 #1` instruction fire and set the
  // value of ret (i.e., %0) to 1.
  __asm__(
  "movw      r6, #0x0;"
  "movt      r6, #0x1;"
  "cmp       r6, #0x0;"
  "movne     %0, #1;"
  : "=r"(ret) /* Outputs */
  : /* Inputs */
  : "r6", "r7" /* Clobbered registers */
  );
  return ret;
}

void _start() {
  test_movt();
}
```
On `movt.unopt.exe`, the failure is at `0x100ec`, i.e., `pop {r6, r7, fp}`
```
armv7l-linux-musleabi-objdump -d -j.text movt.unopt.exe

000100b8 <test_movt>:
   100b8:       e92d08c0        push    {r6, r7, fp}
   100bc:       e28db008        add     fp, sp, #8
   100c0:       e24dd00c        sub     sp, sp, #12
   100c4:       e3a03000        mov     r3, #0
   100c8:       e50b3010        str     r3, [fp, #-16]
   100cc:       e3006000        movw    r6, #0
   100d0:       e3406001        movt    r6, #1
   100d4:       e3560000        cmp     r6, #0
   100d8:       13a03001        movne   r3, #1
   100dc:       e50b3010        str     r3, [fp, #-16]
   100e0:       e51b3010        ldr     r3, [fp, #-16]
   100e4:       e1a00003        mov     r0, r3
   100e8:       e24bd008        sub     sp, fp, #8
   100ec:       e8bd08c0        pop     {r6, r7, fp}
   100f0:       e12fff1e        bx      lr

000100f4 <_start>:
   100f4:       e92d4800        push    {fp, lr}
   100f8:       e28db004        add     fp, sp, #4
   100fc:       ebffffed        bl      100b8 <test_movt>
   10100:       e320f000        nop     {0}
   10104:       e8bd8800        pop     {fp, pc}
```
On `movt.opt.exe`, the failure is at `0x100cc`, i.e., `pop {r6, r7}`:
```
armv7l-linux-musleabi-objdump -d -j.text movt.opt.exe

000100b8 <test_movt>:
   100b8:       e92d00c0        push    {r6, r7}
   100bc:       e3006000        movw    r6, #0
   100c0:       e3406001        movt    r6, #1
   100c4:       e3560000        cmp     r6, #0
   100c8:       13a00001        movne   r0, #1
   100cc:       e8bd00c0        pop     {r6, r7}
   100d0:       e12fff1e        bx      lr

000100d4 <_start>:
   100d4:       e12fff1e        bx      lr
```

langston-barrett:
A simpler version of this test also fails:
```c
int __attribute__((noinline)) test_noop(void) {
  __asm__(
  "nop;"
  : /* Outputs */
  : /* Inputs */
  : "r6", "r7" /* Clobbered registers */
  );
  return 0;
}

int main() {
  return test_noop();
}
```
```
000105b4 <test_noop>:
   105b4:       e92d00c0        push    {r6, r7}
   105b8:       e320f000        nop     {0}
   105bc:       e3a00000        mov     r0, #0
   105c0:       e8bd00c0        pop     {r6, r7}
   105c4:       e12fff1e        bx      lr
```
```
Failed to prove goal: tests/pass/noop.opt.exe:0x105c0: error: in test_noop
```

langston-barrett:
After fixing an off-by-one error (the stack pointer was initially set to one past the end of the stack, instead of at the end), I now see:
```
macaw-aarch32-symbolic tests
  Binary Tests
    Default memory model
      True assertions
        tests/pass/noop.opt.exe: FAIL (0.83s)
          Testing test_noop at 0x105b4 (0.15s)
            tests/Main.hs:249:
            AssertionResult
            expected: SimulationResult Unsat
             but got: SimulationResult Sat

          Use -p '/noop/&&/Default memory model.True assertions.tests\/pass\/noop.opt.exe/' to rerun this test only.
    Lazy memory model
      True assertions
        tests/pass/noop.opt.exe: FAIL (0.48s)
          Testing test_noop at 0x105b4 (0.02s)
            tests/Main.hs:249:
            AssertionResult
            expected: SimulationResult Unsat
             but got: SimulationResult Sat
```
I think what was happening is that the read from `r13` was previously reading at the end of the stack allocation, at offset `0x1000...`. With the fix, the stack pointer gets aligned to something of the form `0xffff..0`, and I believe that the read is now at that offset. This means that it's reading uninitialized bytes (instead of the read failing). I'm not quite sure why that would result in `Sat` here - seems like the return value should still just be concretely 1.

langston-barrett:
Fascinating: it appears that CI passes. No idea why I might be getting `Sat` locally in that case.

RyanGlScott:
Wild. I don't suppose it's due to your local copy not rebuilding things thoroughly?

For what it's worth, the `macaw-aarch32-symbolic` tests also pass when I try them locally.

macaw-aarch32-symbolic/src/Data/Macaw/AArch32/Symbolic/ABI.hs
```diff
@@ -0,0 +1,184 @@
...
+import Lang.Crucible.LLVM.MemModel qualified as MM
+import Lang.Crucible.Simulator qualified as C
+import What4.Interface qualified as WI
+
+-- | Helper, not exported
+ptrRepr :: WI.NatRepr 32
+ptrRepr = WI.knownNat
+
+-- | A pointer to an AArch32 stack
+newtype StackPointer sym = StackPointer { getStackPointer :: MM.LLVMPtr sym 32 }
```
RyanGlScott, line 90: We now have multiple data types named `StackPointer`: one in `macaw-aarch32-symbolic` and another in `macaw-x86-symbolic`. They are similar enough that I wonder if we shouldn't just factor these out into `macaw-symbolic`:

```hs
newtype StackPointer sym w = StackPointer { getStackPointer :: MM.LLVMPtr sym w }
```

Do you agree? (Similar comments apply to `SpilledArgs` and `RetAddr`.)

symbolic/src/Data/Macaw/Symbolic/Stack.hs
```diff
@@ -103,11 +110,127 @@ createArrayStack bak mem slots sz = do
...
+  IO (CLM.LLVMPtr sym w)
+alignStackPointer sym align orig@(CLM.LLVMPointer blk off) = do
+  let alignInt = 2 ^ naturalToInteger (CLD.alignmentToExponent align)
+  if alignInt == 1
+  then pure orig
+  else do
+    -- Because the stack grows down, we can align it by simply ANDing the stack
+    -- pointer with a mask with some amount of 1s followed by some amount of 0s.
+    let mask = BVS.complement ?ptrWidth (BVS.mkBV ?ptrWidth (alignInt - 1))
+    maskBv <- WI.bvLit sym ?ptrWidth mask
```
RyanGlScott, line 155: I think this assumes that the required alignment is a power of two. If so, it would be worth documenting this.

symbolic/src/Data/Macaw/Symbolic/Stack.hs
```diff
@@ -103,11 +110,127 @@ createArrayStack bak mem slots sz = do
   (base, mem1) <- mallocStack bak mem sz
   mem2 <- CLM.doArrayStore bak mem1 base stackAlign arr sz
 
-  end <- CLM.doPtrAddOffset bak mem2 base sz
+  off <- WI.bvSub sym sz =<< WI.bvOne sym ?ptrWidth
+  end <- CLM.doPtrAddOffset bak mem2 base off
```
RyanGlScott, line 114: Can you leave a comment explaining what is going on here? This ended up being surprisingly subtle, and I'll admit that I'm not entirely sure why this part is necessary, despite knowing that it was important for fixing the test suite.

RyanGlScott (approved):

macaw-aarch32-symbolic/src/Data/Macaw/AArch32/Symbolic/ABI.hs
langston-barrett: I actually think it's nice to have these as separate types, even though they have similar representations. They both contain pointers, but those pointers have different constraints, and having separate types prevents confusion between them. For example, it would be wrong to pass a pointer returned from `Data.Macaw.AArch32.Symbolic.ABI.stackPointerReg` into `Data.Macaw.X86.Symbolic.ABI.SysV.pushStackFrame`, because the ABI constraints for AArch32 and SysV are different. Perhaps more realistically, consider what would happen if we added support for the Microsoft x86_64 ABI as well. Then, we wouldn't want to confuse pointers to a Microsoft-compatible stack with pointers to a SysV-compatible stack.

symbolic/src/Data/Macaw/Symbolic/Stack.hs
langston-barrett: Isn't alignment always a power of 2? Note the definition of `alignInt`:
```haskell
  let alignInt = 2 ^ naturalToInteger (CLD.alignmentToExponent align)
```
The Haddocks on `Alignment` state:

> An Alignment represents a number of bytes that must be a power of two.

RyanGlScott: Oops, quite right. Please ignore me.

PR #446, opened 2024-10-09
macaw-dump: A CLI for printing internal data structures

langston-barrett:
A simple library and tiny wrappers for different architectures, intended for debugging purposes. The only current capability is to run code discovery on a set of symbols in a binary (or all of them if none are specified), and pretty-print the resulting Macaw or Crucible CFGs.

Needs:

- [x] Definition of a CLI in the library (likely using `optparse-applicative`)
- [x] Binaries for each architecture
    - [x] AArch32 
    - [x] PPC32
    - [ ] PPC64
    - [x] RISCV
    - [x] x86_64
- [x] PLT stub dumper


langston-barrett:
Not sure what this error from specifically GHC 9.4 is all about:
```
tools/Dump.hs:35:24: error:
Error:     • Could not deduce (macaw-base-0.3.15.6:Data.Macaw.CFG.AssignRhs.RegAddrWidth
                          (macaw-base-0.3.15.6:Data.Macaw.CFG.AssignRhs.ArchReg
                             semmc-aarch[32](https://github.com/GaloisInc/macaw/actions/runs/11258785738/job/31306208454?pr=446#step:20:33)-0.1.0.0:SemMC.Architecture.AArch32.AArch32)
                        ~ 32)
        arising from a use of ‘MD.runDiscovery’
```

RyanGlScott:
Sigh. This is https://gitlab.haskell.org/ghc/ghc/-/issues/16234. The good news is that this is fixed in GHC 9.6 or later, but the bad news is that we'll have to live with the bug until we drop GHC 9.4 support. Happily, it is straightforward to work around the bug:

```diff
diff --git a/macaw-aarch32-symbolic/tools/Dump.hs b/macaw-aarch32-symbolic/tools/Dump.hs
index a11a6ca5..09b0bf03 100644
--- a/macaw-aarch32-symbolic/tools/Dump.hs
+++ b/macaw-aarch32-symbolic/tools/Dump.hs
@@ -3,6 +3,15 @@
 
 module Main (main) where
 
+-- Sometimes, GHC 9.4 is unable to find instances of RegAddrWidth that are
+-- available by way of transitive module imports. The only reliable way of
+-- preventing this issue that I've found is to import the defining module for
+-- the instances (Data.Macaw.ARM.ARMReg) directly. See
+-- https://gitlab.haskell.org/ghc/ghc/-/issues/16234 for the upstream GHC issue.
+--
+-- This issue does not affect GHC 9.6 or later, so when we drop support for GHC
+-- 9.4, we can remove the redundant import below.
+import Data.Macaw.ARM.ARMReg ()
 import Data.Macaw.AArch32.Symbolic ()
 import Data.Macaw.ARM qualified as MA
 import Data.Macaw.Dump qualified as MD
```

langston-barrett:
@RyanGlScott Can you take a quick look and see if you like this functionality, and if so, I'll add the other architectures?

macaw-dump/src/Data/Macaw/Dump/Discover.hs
```diff
@@ -0,0 +1,160 @@
...
+
+
+-- | Convert machine addresses into What4 positions.
+--
+-- When possible, we map to the structured 'WPL.BinaryPos' type. However, some
+-- 'MM.MemSegmentOff' cannot be mapped to an absolute position (e.g., some
+-- addresses from shared libraries are in non-trivial segments). In those cases,
+-- we map to the unstructured 'WPL.Others' with a sufficiently descriptive string.
+--
+-- TODO: import from `Testing`
```
RyanGlScott, line 60: Does this TODO still need to be resolved?

macaw-dump/src/Data/Macaw/Dump/Discover.hs
```diff
@@ -0,0 +1,160 @@
...
+data DiscoverException = ELFResolutionError String
+  deriving Show
+
+instance X.Exception DiscoverException
+
+
+-- | Convert machine addresses into What4 positions.
+--
+-- When possible, we map to the structured 'WPL.BinaryPos' type. However, some
+-- 'MM.MemSegmentOff' cannot be mapped to an absolute position (e.g., some
```
RyanGlScott, line 56:
```suggestion
-- When possible, we map to the structured 'WPL.BinaryPos' type. However, some
-- 'MM.MemSegmentOff' values cannot be mapped to an absolute position (e.g., some
```

macaw-dump/src/Data/Macaw/Dump/Discover.hs
```diff
@@ -0,0 +1,160 @@
...
+    IO.print (PP.pretty info)
+    Monad.when printCrucible $ do
+      let pos = posFn (Text.pack path)
+      let funName =
+            WF.functionNameFromText $
+              Text.decodeUtf8With Text.lenientDecode $
+                MD.discoveredFunName info
+      CCC.SomeCFG ssa <-
+        MS.mkFunCFG (MS.archFunctions archVals) halloc funName pos info
+      IO.print (CCC.ppCFG' True (CAP.postdomInfo ssa) ssa)
```
RyanGlScott, line 128: Does it make sense to include some kind of headers to make it obvious which of the two `print`ed values correspond to which CFGs?

macaw-dump/src/Data/Macaw/Dump/Discover.hs
```diff
@@ -0,0 +1,160 @@
...
+  , discSymbols :: [BS.ByteString]
+    -- Options
+  , discPrintCrucible :: Bool
+  }
+
+discoverConfig :: Opt.Parser DiscoverConfig
+discoverConfig =
+  DiscoverConfig
+  <$> MDCU.binOpt
+  <*> Opt.many (Opt.strArgument (Opt.help "function name" <> Opt.metavar "SYMBOL"))
```
RyanGlScott, line 142: It might be worth mentioning in the help text that the function names must be UTF-8–encoded.

macaw-dump/src/Data/Macaw/Dump/Plt.hs
```diff
@@ -0,0 +1,46 @@
...
+import Data.Macaw.CFG qualified as MC
+import Data.Macaw.Dump.CLIUtils qualified as MDCU
+import Data.Macaw.Memory.ElfLoader.PLTStubs as MMEP
+import Data.Macaw.Memory.LoadCommon qualified as MML
+import Data.Macaw.Memory qualified as MM
+import Options.Applicative qualified as Opt
+import Prettyprinter as PP
+import System.IO qualified as IO
+
+data PltConfig = PltConfig
```
RyanGlScott, line 24:
```suggestion
newtype PltConfig = PltConfig
```

RyanGlScott (commented):
This looks great! I think there are definitely ways that we could make the results more configurable (e.g., fine-tuning how code discovery is performed, controlling the base address, etc.), but this would be a good foundation to build on later.

macaw-dump/src/Data/Macaw/Dump.hs
```diff
@@ -0,0 +1,35 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE TypeOperators #-}
+{-# OPTIONS_GHC -Wno-simplifiable-class-constraints #-}
```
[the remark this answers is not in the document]
langston-barrett: Well, without suppressing it I see:
```haskell
src/Data/Macaw/Dump.hs:20:3: error: [-Wsimplifiable-class-constraints, Werror=simplifiable-class-constraints]
    • The constraint ‘CCE.IsSyntaxExtension (MS.MacawExt arch)’ matches
        instance macaw-symbolic-0.0.1:Data.Macaw.Symbolic.CrucGen.MacawArchConstraints
                   arch =>
                 CCE.IsSyntaxExtension (MS.MacawExt arch)
          -- Defined in ‘macaw-symbolic-0.0.1:Data.Macaw.Symbolic.CrucGen’
      This makes type inference for inner bindings fragile;
        either use MonoLocalBinds, or simplify it using the instance
    • In the type signature:
        main :: (MS.GenArchInfo MS.LLVMMemory arch,

                 CCE.IsSyntaxExtension (MS.MacawExt arch),
                 MC.ArchConstraints arch,
                 MC.ArchAddrWidth arch ~ EE.RelocationWidth reloc,
                 EE.IsRelocationType reloc) =>
                MAI.ArchitectureInfo arch
                -> MS.GenArchVals mem arch -> MMEP.PLTStubInfo reloc -> IO ()
   |
20 |   ( MS.GenArchInfo MS.LLVMMemory arch
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```
But if I remove the `IsSyntaxExtension` constraint, I then get:
```
src/Data/Macaw/Dump.hs:32:32: error: [GHC-39999]
    • Could not deduce ‘Data.Parameterized.TraversableFC.TraversableFC
                          (macaw-symbolic-0.0.1:Data.Macaw.Symbolic.CrucGen.MacawArchStmtExtension
                             arch)’
        arising from a use of ‘MDD.discover’
      from the context: (MS.GenArchInfo MS.LLVMMemory arch,
                         MC.ArchConstraints arch,
                         MC.ArchAddrWidth arch ~ EE.RelocationWidth reloc,
                         EE.IsRelocationType reloc)
        bound by the type signature for:
                   main :: forall arch reloc mem.
                           (MS.GenArchInfo MS.LLVMMemory arch, MC.ArchConstraints arch,
                            MC.ArchAddrWidth arch ~ EE.RelocationWidth reloc,
                            EE.IsRelocationType reloc) =>
                           MAI.ArchitectureInfo arch
                           -> MS.GenArchVals mem arch -> MMEP.PLTStubInfo reloc -> IO ()
        at src/Data/Macaw/Dump.hs:(19,1)-(28,7)
    • In the expression: MDD.discover archInfo archVals cfg
      In a case alternative:
          MDC.CommandDiscover cfg -> MDD.discover archInfo archVals cfg
      In a stmt of a 'do' block:
        case MDC.cliCommand cli of
          MDC.CommandDiscover cfg -> MDD.discover archInfo archVals cfg
          MDC.CommandPlt cfg -> MDP.plt archInfo pltStubInfo cfg
   |
32 |     MDC.CommandDiscover cfg -> MDD.discover archInfo archVals cfg
   |                                ^^^^^^^^^^^^
```
I get a different error if I remove `ArchConstraints`.

macaw-dump/src/Data/Macaw/Dump/CLI.hs
```diff
@@ -0,0 +1,59 @@
...
+      "discover"
+      (Opt.info (CommandDiscover <$> MDD.discoverConfig) (Opt.progDesc "Perform code discovery and print CFGs"))
+
+  cmdPlt :: Opt.Mod Opt.CommandFields Command
+  cmdPlt = do
+    Opt.command
+      "plt"
+      (Opt.info (CommandPlt <$> MDP.pltConfig) (Opt.progDesc "Display PLT stubs"))
+
+data Cli = Cli
```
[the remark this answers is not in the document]
langston-barrett: I made this a `data` in anticipation of adding other fields - Do you think that's just not worth it, and we should keep it as a `newtype` until a need for more fields arises? My only nervousness is that this has an exported constructor, so it being a `newtype` is in some ways part of its API (via e.g., `coerce`). Thus, changing it to `data` later could concievably lead to breakage.

macaw-dump/src/Data/Macaw/Dump/Discover.hs
langston-barrett: Eh, it's a pretty small function, I'm not too worried about this level of duplication on second thought :shrug:

macaw-dump/src/Data/Macaw/Dump.hs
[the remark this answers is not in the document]
RyanGlScott: Huh, I'm surprised that we don't export `MacawArchConstraints`, as it seems like the most direct way to write this would be replacing the `IsSyntaxExtension (MS.MacawExt arch)` constraint with `MacawArchConstraints arch`. But failing that, I'd favor enabling `MonoLocalBinds` rather than disabling the warning wholesale.

macaw-dump/src/Data/Macaw/Dump/CLI.hs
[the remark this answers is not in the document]
RyanGlScott: Fair enough. I'm fine with keeping this as a data type if it's destined to add more fields soon.

macaw-dump/src/Data/Macaw/Dump/Discover.hs
langston-barrett: What makes you say that? They are stored as `ByteString`s.

RyanGlScott: Yes, but `displayCfgs` decodes them to `Text` under the assumption that they're UTF-8–encoded.

langston-barrett: Yes, but it does so with [`lenientDecode`](https://hackage.haskell.org/package/text-2.1.2/docs/Data-Text-Encoding-Error.html#v:lenientDecode), so I don't think I would say they *must* be UTF-8 encoded.

RyanGlScott: > I don't think I would say they must be UTF-8 encoded.

On the other hand, you'll get very surprising results if you give it a UTF-16–encoded string like `roșu`. (It won't cause your program to crash, but it might make you scratch your head wondering why `roșu` turned into `rou`.)

In any case, this isn't that big of a deal, but I usually like to draw attention to this kinds of potential pitfalls.

langston-barrett: Totally - I'm trying to grok whether this is indeed a pitfall or if it's actually OK. Very happy to document it if it does end up being confusing. The docs for `lenientDecode` say:

> Replace an invalid input byte with the Unicode replacement character U+FFFD.

[Wikipedia says](https://en.wikipedia.org/wiki/Specials_(Unicode_block)):

> U+FFFD � replacement character used to replace an unknown, unrecognised, or unrepresentable character

To me, these together indicate that your example `roșu`, if it were UTF-16 encoded, would print as `ro�u` or the like, which IMO reasonably clearly indicates an encoding error. Or actually, since UTF-16 isn't even ASCII-compatible, perhaps it would just print as `����`.

Another option would be to print a warning in the case that the function name actually wasn't completely valid UTF-8.

langston-barrett: Added a note that symbol names are expected but not required to be UTF-8.

RyanGlScott: It likely depends on what locale you're using when printing, but for me, the following happens:

```
> Data.Text.IO.putStrLn $ decodeUtf8With lenientDecode $ encodeUtf16LE "roșu"
rou
```

> Another option would be to print a warning in the case that the function name actually wasn't completely valid UTF-8.

No need to go out of your way for this—I think just leaving a comment somewhere near the place where they're parsed is fine.

langston-barrett: > It likely depends on what locale you're using when printing, but for me, the following happens:

TIL, thanks for the example! I will try to remember to be more careful about this in the future.

macaw-dump/src/Data/Macaw/Dump/Discover.hs
```diff
@@ -0,0 +1,163 @@
...
+
+instance X.Exception DiscoverException
+
+
+-- | Convert machine addresses into What4 positions.
+--
+-- When possible, we map to the structured 'WPL.BinaryPos' type. However,
+-- some 'MM.MemSegmentOff' values cannot be mapped to an absolute position
+-- (e.g., some addresses from shared libraries are in non-trivial segments).
+-- In those cases, we map to the unstructured 'WPL.Others' with a sufficiently
```
RyanGlScott, line 58: I think this is meant to be:

```suggestion
-- In those cases, we map to the unstructured 'WPL.OtherPos' with a sufficiently
```

macaw-dump/src/Data/Macaw/Dump/Discover.hs
```diff
@@ -0,0 +1,163 @@
...
+  MS.GenArchVals mem arch ->
+  -- | Also print Crucible CFG?
+  Bool ->
+  IO ()
+displayCfgs path discState archVals printCrucible = do
+  let funInfos = discState Lens.^. MD.funInfo
+  halloc <- CFH.newHandleAllocator
+  F.for_ (Map.toList funInfos) $ \(_addr, Some info) -> do
+    Monad.when printCrucible $ do
+      IO.putStrLn "== Macaw =="
```
RyanGlScott, line 119: Is this `putStrLn` meant to be guarded behind the `printCrucible` option? What's more, the printing of the `== Macaw ==` header is guarded behind it, but not the printing of Macaw-related information on the line directly below...

symbolic/src/Data/Macaw/Symbolic/Testing.hs
```diff
@@ -113,10 +113,11 @@ instance X.Exception TestingException
...
 -- | Convert machine addresses into Crucible positions.
 --
--- When possible, we map to the structured 'WPL.BinaryPos' type. However, some
--- 'MM.MemSegmentOff' cannot be mapped to an absolute position (e.g., some
--- addresses from shared libraries are in non-trivial segments). In those cases,
--- we map to the unstructured 'WPL.Others' with a sufficiently descriptive string.
+-- When possible, we map to the structured 'WPL.BinaryPos' type. However,
+-- some 'MM.MemSegmentOff' values cannot be mapped to an absolute position
+-- (e.g., some addresses from shared libraries are in non-trivial segments).
+-- In those cases, we map to the unstructured 'WPL.Others' with a sufficiently
```
RyanGlScott, line 119: Similarly:

```suggestion
-- (e.g., some addresses from shared libraries are in non-trivial segments).
-- In those cases, we map to the unstructured 'WPL.OtherPos' with a sufficiently
```

macaw-aarch32-symbolic/tools/Dump.hs
```diff
@@ -0,0 +1,27 @@
...
+import Data.Macaw.Symbolic qualified as MS
+import Data.Proxy (Proxy(..))
+
+main :: IO ()
+main = do
+  archVals <-
+    case MS.archVals (Proxy @MA.ARM) Nothing of
+      Just archVals -> pure archVals
+      Nothing -> error "impossible"
+  MD.main MA.arm_linux_info archVals MA.armPLTStubInfo
```
RyanGlScott, line 27: There are instantiations of `macaw-dump` for all architectures except RISC-V. Is this intended?

macaw-ppc-symbolic/tools/Dump.hs
```diff
@@ -0,0 +1,23 @@
...
+
+main :: IO ()
+main = do
+  archVals <-
+    case MS.archVals (Proxy @PPC.PPC32) Nothing of
+      Just archVals -> pure archVals
+      Nothing -> error "impossible"
+  let pltStubInfo :: MMEP.PLTStubInfo EE.PPC32_RelocationType
+      pltStubInfo = error "PLT stub discovery is not supported on PPC"
+  MD.main PPC.ppc32_linux_info archVals pltStubInfo
```
RyanGlScott, line 23: Is there a reason to hard-code this to 32-bit PPC here?

macaw-dump/src/Data/Macaw/Dump/Discover.hs
langston-barrett: Yeah, my reasoning was that there's no reason to print a disambiguating header when only printing the Macaw CFG.

macaw-aarch32-symbolic/tools/Dump.hs
langston-barrett: Yeah, AFAIK there's no RISCV-symbolic support, so there's no way to use `Dump.main`, which requires things like `MS.GenArchInfo`.

RyanGlScott: > AFAIK there's no RISCV-symbolic support,

[There isn't?](https://github.com/GaloisInc/macaw/tree/master/macaw-riscv-symbolic)

macaw-dump/src/Data/Macaw/Dump/Discover.hs
RyanGlScott: Ah, that makes sense.

macaw-ppc-symbolic/tools/Dump.hs
langston-barrett: Just that it would be hard with the current architecture to select between them at the CLI... See most recent commit/message. I made separate executables 32- and 64-bit for now.

RyanGlScott: > What would you think about deferring PPC64 support to an issue?

I'm fine with that.

RyanGlScott (approved):

PR #452, opened 2024-11-22
`macaw-symbolic`: Expose memory model read/write functionality

RyanGlScott:
This exposes the default implementations of `MacawReadMem`, `MacawWriteMem`, and friends as top-level functions in `Data.Macaw.Symbolic` and exports them. This makes it possible for downstream clients to perform memory reads and writes in a way that respects the `macaw-symbolic` memory model.


langston-barrett (approved):
I'm assuming that the bodies of the functions were just copy/pasted without signficant changes. If so, LGTM! If not, I can do a more thorough review. Thanks a bunch for the docs!

RyanGlScott:
> I'm assuming that the bodies of the functions were just copy/pasted without signficant changes.

Yes, that is correct.

PR #456, opened 2024-12-05
`macaw-riscv-symbolic`: `Index`es for registers

RyanGlScott:
Like #445, but for RISC-V.


macaw-riscv-symbolic/src/Data/Macaw/RISCV/Symbolic/Regs.hs
```diff
@@ -0,0 +1,1044 @@
+{-|
+Copyright        : (c) Galois, Inc 2024
+Maintainer       : Langston Barrett <langston@galois.com>
```
langston-barrett, line 3: Maybe this should be you? :slightly_smiling_face:

RyanGlScott:
While I'm not entirely happy with the compile times in `Data.Macaw.RISCV.Symbolic.Regs` (which currently takes just over 10 seconds to compile), none of my attempts at rewriting it proved very successful. As such, I'll just accept the cost.

PR #461, opened 2025-02-10
x86-symbolic: Improve div-by-zero error messages

langston-barrett:
- Phrase them in terms of abstractions users can understand (instructions, rather than our internal representation of x86_64 semantics)
- Don't negate what has happened

### DIFF
--- a/x86_symbolic/src/Data/Macaw/X86/Crucible.hs
+++ b/x86_symbolic/src/Data/Macaw/X86/Crucible.hs
@@ -730,13 +730,9 @@ getDenominator dw sym op macawDenom = do
      assert bak denNotZero (C.AssertFailureSimError errMsg "")
   pure den
 
--- | Performs a simple unsigned division operation.
+-- | Performs unsigned division, i.e., the @div@ instruction.
 --
--- The x86 numerator is twice the size as the denominator.
---
--- This function is only reponsible for the dividend (not any
--- remainder--see uRem for that), and any divide-by-zero exception was
--- already handled via an Assert.
+-- Asserts that the denominator is not zero and the quotient doesn't overflow.
 uDivRem :: forall sym bak w
         .  (IsSymBackend sym bak)
         => Sym sym bak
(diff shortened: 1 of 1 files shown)


x86_symbolic/src/Data/Macaw/X86/Crucible.hs
```diff
@@ -755,7 +756,7 @@ uDivRem sym repsz macawNum1 macawNum2 macawDenom =
     -- Get numerator
     numExt <- getNumerator dw sym macawNum1 macawNum2
     -- Get denominator
-    den <- getDenominator dw sym macawDenom
+    den <- getDenominator dw sym "div" macawDenom
```
RyanGlScott, line 759: There's something a bit funny going on here. The error message specifically mentions the `div` instruction, but the name of this function (`uDivRem`) describes itself as performing a general unsigned division operation. Given that we are now encoding the name of the instruction itself in the error message, I wonder if we should just update the Haddocks to say that this is meant to power the `div` instruction in particular.

(Relatedly, the Haddocks claim that it does not handle remainders, but this appears to be out of date with the implementation.)

x86_symbolic/src/Data/Macaw/X86/Crucible.hs
```diff
@@ -795,7 +796,7 @@ sDivRem sym repsz macawNum1 macawNum2 macawDenom =
     -- Get numerator
     numExt <- getNumerator dw sym macawNum1 macawNum2
     -- Get denominator
-    den <- getDenominator dw sym macawDenom
+    den <- getDenominator dw sym "idiv" macawDenom
```
RyanGlScott, line 799: Similarly, I think we should mention that this function (`sDivRem`) is specifically meant to power the `idiv` instruction.

RyanGlScott (approved):
LGTM, aside from a request for Haddock clarification.

langston-barrett:
Good call on the Haddocks, updated!

Issue #283, opened 2022-04-29
Untangle the `macaw-loader` submodule knot 


RyanGlScott:
Currently the `macaw` repo depends on the [`macaw-loader`](https://github.com/GaloisInc/macaw-loader) repo as a submodule dependency. However, the `macaw-loader` repo _also_ depends on the `macaw` repo as a submodule. This makes coordinating submodule bumps between the repos challenging, as they sometimes have to be updated simultaneously.

I'd like to propose that we find a way to make this situation more tolerable. Some possible ideas for doing so:

1. Cut the submodule dependency knot by removing `macaw`'s dependency on `macaw-loader`, thereby removing the `macaw-loader` submodule dependency from `macaw`. This would require changing some code in `macaw` to avoid the use of `macaw-loader`, however. In some cases, this would impact the public-facing API, such as in [`macaw-ppc`](https://github.com/GaloisInc/macaw/blob/2c85dce18e2c38cef160ebe13de95a678806bc2f/macaw-ppc/src/Data/Macaw/PPC.hs#L93) and [`macaw-refinement`](https://github.com/GaloisInc/macaw/blob/bbc0b6a40a2b99875c45fe137c97cc6c9e142f7d/refinement/src/Data/Macaw/Refinement/SymbolicExecution.hs#L116).
2. Bring the `macaw-loader` packages into the `macaw` repo. This would involve some amount of churn for downstream libraries, but it could be done.
3. Something else?

travitch:
I'm happy to merge it in. Id like to think about the API a bit, but maybe we can just mark it unstable and refine it over time

kquick:
I also like option 2: merging in macaw-loader.  IIRC, it wasn't originally a circular dependency until we merged some of the other macaw repos together.

RyanGlScott:
Another unfortunate consequence of doing this is that `git submodule update --init --recursive` will go into an infinite loop, as reported in https://github.com/GaloisInc/saw-script/issues/2065.

PR #470, opened 2025-02-24
Merge the `macaw-loader` repo into `macaw`

RyanGlScott:
This merges the entire `git` history of the `macaw-loader` repo into `macaw`, thereby incorporating the `macaw-loader-*` packages as sub-packages of the `macaw` repo. This breaks the knot `macaw` and `macaw-loader` repos, thereby fixing #283.


RyanGlScott:
I have spoken with @sauclovian-g about this, and he is on board with trying to land this before a new release of SAW (a tool which would be impacted by the `macaw-loader` submodule switch). Some amount of downstream churn will be inevitable, but I think the payoff of breaking the submodule knot will be very much worth it.

I've also added reviewers on other projects who will be impacted to get a variety of different opinions on this.

langston-barrett (approved):
I agree that the payoff is worth it, and it makes sense to merge these repos.

macaw-loader-aarch32/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright (c) 2018, Galois, Inc.
```
langston-barrett, line 1: Consider updating copyright year

macaw-loader-aarch32/Setup.hs
```diff
@@ -0,0 +1,2 @@
+import Distribution.Simple
+main = defaultMain
```
langston-barrett, line 2: Is this necessary?

macaw-loader-aarch32/macaw-loader-aarch32.cabal
```diff
@@ -0,0 +1,28 @@
+name:                macaw-loader-aarch32
+version:             0.1.0.0
+synopsis:            Library to load ARM binary images and get macaw Memory and entry points.
+-- description:
+homepage:            https://github.com/GaloisInc/macaw-loader/macaw-loader-arm
+license:             BSD3
+license-file:        LICENSE
+author:              Galois, Inc.
+maintainer:          kquick@galois.com
+copyright:           2018, Galois Inc.
```
langston-barrett, line 10: Consider updating copyright year

langston-barrett (commented):
Just a few nits. The copyright year ones apply to other Cabal and LICENSE files as well

README.org
```diff
@@ -0,0 +1,17 @@
+This library provides a uniform interface to load a binary (e.g. in
```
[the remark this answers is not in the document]
kquick: Both are rendered fine by github, and both are fairly understandable, so if there was a desire to do this it would probably be better in a different PR.  My recommendation would be to not worry about it though.

macaw-loader-aarch32/LICENSE
kquick: I'd also recommend a separate PR for this.  Although advice from our legal dept. is that churning these dates isn't necessary to maintain copyright.

README.org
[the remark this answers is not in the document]
RyanGlScott: I was just about to comment the same thing. What's more, there are _already_ subpackages that use `README.org` files (`macaw-{aarch32,riscv,ppc,x86}-symbolic`, `macaw-refinement`, and `macaw-symbolic` all do this). As such, I'll choose to leave this alone for now.

kquick (approved):
I am not opposed to this merger.

macaw-loader-aarch32/LICENSE
RyanGlScott: OK. For what it's worth, I've pushed a separate commit (https://github.com/GaloisInc/macaw/pull/470/commits/a0c7f7a4598a0327961f2561468514e5ebecd64a) that updates the copyright years, but only for the `macaw-loader` packages (since those are the ones that are actually touched in this PR). If you'd prefer, I could revert this and defer it to a more general PR.

langston-barrett: > Although advice from our legal dept. is that churning these dates isn't necessary to maintain copyright.

TIL, thanks! That should save us all some time and mundanity :grin:

README.org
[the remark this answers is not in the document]
langston-barrett: Whoops, my bad for not checking to see that the current documentation uses a mix of Markdown and Org. Good catch!

macaw-loader-aarch32/LICENSE
kquick: Not a big change and it's in a separate commit (thus bisectable), so don't worry about backing it out now.

sauclovian-g (approved):
Definitely good to get this problem fixed.

PR #478, opened 2025-03-31
Remove nonexistent extra files from .cabals to fix sdist

glguy:


sauclovian-g (approved):
At least for the LICENSE though it might be better to copy one in, especially if most or all of the other subpackages have LICENSEs.

Issue #482, opened 2025-04-16
`macaw-loader` crashes on binaries with no entry point


RyanGlScott:
Per the [ELF specification's documentation](https://refspecs.linuxbase.org/elf/gabi4+/ch4.eheader.html) of the `e_entry` field (regarding ELF header entry points):

> `e_entry`
>
> This member gives the virtual address to which the system first transfers control, thus starting the process. If the file has no associated entry point, this member holds zero. 

As this indicates, an ELF binary is not strictly required to have an entry point. Despite this, `macaw-loader` will crash if you use it to load a binary without an entry point. As an example, take a simple C program like this one:

```c
int main(void) {
  return 0;
}
```

Then compile it to a statically linked x86-64 binary using [`musl-gcc`](https://musl.cc/), telling the linker to set the entrypoint to a non-existent address using `-Wl,--entry=0`:

```
$ musl-gcc test.c -static -Wl,--entry=0 -o test.exe
```

Now attempt to load `test.exe` using the following harness written on top of `macaw-loader-x86`:

```hs
-- Bug.hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
{-# OPTIONS_GHC -Wall #-}
module Main (main) where

import qualified Data.ByteString as BS
import qualified Data.ElfEdit as EE
import qualified System.Environment as SE

import qualified Data.Macaw.BinaryLoader as DMB
import qualified Data.Macaw.Memory.LoadCommon as MML

import Data.Macaw.BinaryLoader.X86 ()
import qualified Data.Macaw.X86 as MX

main :: IO ()
main = do
  args <- SE.getArgs
  elfPath <-
    case args of
      [elfPath] -> pure elfPath
      _ -> fail "Usage: <prog> <path-to-x86_64-ELF-binary>"
  bytes <- BS.readFile elfPath
  case EE.decodeElfHeaderInfo bytes of
    Left (_off, msg) -> fail msg
    Right (EE.SomeElf e) ->
      case EE.headerClass (EE.header e) of
        EE.ELFCLASS32 ->
          fail "Not an x86_64 binary"
        EE.ELFCLASS64 -> do
          lb <- DMB.loadBinary @MX.X86_64 options e
          entryPoints <- DMB.entryPoints lb
          print entryPoints
  where
    options = MML.LoadOptions { MML.loadOffset = Just 0 }
```

Now observe that this harness program crashes:

```
$ ghc Bug.hs
[...]

PR #483, opened 2025-04-16
`macaw-dump`: Add `entry-points` subcommand

RyanGlScott:
This prints the entry point addresses computed by `macaw-loader`. This proves useful to debug issues like the one observed in https://github.com/GaloisInc/macaw/issues/482.


langston-barrett (approved):
Nice, glad you're finding `macaw-dump` useful!

PR #484, opened 2025-04-17
`macaw-loader`: Make `entryPoints` return a possibly empty list

RyanGlScott:
It is not guaranteed that an ELF file's `e_entry` value will always correspond to a valid entry point address, which means that we cannot guarantee that the list of entry points that `macaw-loader`'s `entryPoints` computes will always be non-empty. This patch changes the type of `entryPoints` to return `[]` instead of `NonEmpty` and makes it so that failing to resolve the `e_entry` address in `entryPoints` does not throw an exception.

Fixes https://github.com/GaloisInc/macaw/issues/482.


RyanGlScott:
I expect the downstream breakage from this change to be fairly minimal, although I do spot [one place in `grease`](https://github.com/GaloisInc/grease/blob/3fc2fde2137c6a2cdd92fe1dbadf7d15a3d4f21e/grease/src/Grease/Macaw/Load.hs#L144) that will need to be updated.

langston-barrett (approved):

PR #486, opened 2025-04-17
`Data.Macaw.Memory.ElfLoader`: Replace error with `ShdrPhdrOverlap` warning

### LINKED ISSUE #485: `Data.Macaw.Memory.ElfLoader`: Downgrade program/section header overlap from an error to a warning
I have an ELF binary with a custom `._TEXT` section header that spans multiple `LOAD` program headers. I believe that this is a perfectly fine thing for ELF binaries to do, as nothing in [the ELF specification](https://refspecs.linuxbase.org/elf/gabi4+/ch5.pheader.html) explicitly prevents section headers from overlapping with program headers in this way. Despite this, `macaw`'s ELF loader is not very happy with the binary, as it produces the following error when it attempts to load it:

```
Found section header that overlaps with program header.
  error, called at src/Data/Macaw/Memory/ElfLoader.hs:1454:13 in macaw-base-0.3.15.6-inplace:Data.Macaw.Memory.ElfLoader
```

This error arises from here:

https://github.com/GaloisInc/macaw/blob/28c08f20a832e5c98dc244445350a9c7f009563a/base/src/Data/Macaw/Memory/ElfLoader.hs#L1456-L1459

I would argue that this should be a warning at most. Specifically, I think that we could make a tweak that looks something like this:

<details>

```
diff --git a/base/src/Data/Macaw/Memory/ElfLoader.hs b/base/src/Data/Macaw/Memory/ElfLoader.hs
index ac3963d3..491b08e5 100644
--- a/base/src/Data/Macaw/Memory/ElfLoader.hs
+++ b/base/src/Data/Macaw/Memory/ElfLoader.hs
@@ -195,6 +195,8 @@ data MemLoadWarning
     -- ^ Multiple relocations at the given offset
   | IgnoreRelocation !RelocationError
     -- ^ @IgnoreRelocation err@ warns we ignored a relocation.
+  | ShdrPhdrOverlap
+    -- ^ A section header overlaps with a program header.

 ppSymbol :: SymbolName -> String
 ppSymbol "" = "unnamed symbol"
@@ -250,6 +252,8 @@ instance Show MemLoadWarning where
     "Multiple relocations modify " ++ showHex addr "."
   show (IgnoreRelocation err) =
     show err
+  show ShdrPhdrOverlap =
+    "Found section header that overlaps with program header."

 data MemLoaderState w = MLS { _mlsMemory :: !(Memory w)
                             , mlsEndianness :: !Endianness
@@ -1449,16 +1453,17 @@ insertElfSegment regIdx addrOff shdrMap contents relocMap phdr = do
     let l = IMap.toList $ IMap.intersecting shdrMap (IntervalCO phdrOffset phdrEnd)
     forM_ l $ \(i, elfIdx) -> do
       case i of
-        IntervalCO shdr_start _ -> do
-          when (phdrOffset > shdr_start) $ do
-            error "Found section header that overlaps with program header."
-          let sec_offset = fromIntegral $ shdr_start - phdrOffset
-          addr <- case resolveSegmentOff seg sec_offset of
-                    Just addr -> pure addr
-                    Nothing -> error $ "insertElfSegment: Failed to resolve segment offset at "
-                                    ++ show sec_offset
-          mlsMemory   %= memBindSectionIndex elfIdx addr
-          mlsIndexMap %= Map.insert elfIdx addr
+        IntervalCO shdr_start _
+          | phdrOffset > shdr_start ->
+              addWarning ShdrPhdrOverlap
+          | otherwise -> do
+              let sec_offset = fromIntegral $ shdr_start - phdrOffset
+              addr <- case resolveSegmentOff seg sec_offset of
+                        Just addr -> pure addr
+                        Nothing -> error $ "insertElfSegment: Failed to resolve segment offset at "
+                                        ++ show sec_offset
+              mlsMemory   %= memBindSectionIndex elfIdx addr
+              mlsIndexMap %= Map.insert elfIdx addr
         _ -> error "Unexpected shdr interval"

 -- | Load an elf file into memory by parsing segments.
```
</details>

This would mean that any section headers that overlap a program header would be dropped. This should be fine, I think, as if a section header spans multiple program headers, then one of them (the one that doesn't fall afoul of `phdrOffset > shdr_start`) will still be included, so it will still show up in the loaded ELF binary. As such, this change should allow strictly more ELF binaries to be loaded.

RyanGlScott:
This downgrades an error in `insertElfSegment` to a warning, represented by the new `ShdrPhdrOverlap` data constructor of `MemLoadWarning`. This allows `macaw-base`'s ELF loader to load binaries where a section header overlaps a program header, which is legal (albeit uncommon) in ELF.

Fixes #485.

### DIFF
--- a/base/src/Data/Macaw/Memory/ElfLoader.hs
+++ b/base/src/Data/Macaw/Memory/ElfLoader.hs
@@ -195,6 +195,8 @@ data MemLoadWarning
     -- ^ Multiple relocations at the given offset
   | IgnoreRelocation !RelocationError
     -- ^ @IgnoreRelocation err@ warns we ignored a relocation.
+  | ShdrPhdrOverlap
+    -- ^ A section header overlaps with a program header.
 
 ppSymbol :: SymbolName -> String
 ppSymbol "" = "unnamed symbol"
@@ -250,6 +252,8 @@ instance Show MemLoadWarning where
     "Multiple relocations modify " ++ showHex addr "."
   show (IgnoreRelocation err) =
     show err
+  show ShdrPhdrOverlap =
+    "Found section header that overlaps with program header."
 
 data MemLoaderState w = MLS { _mlsMemory :: !(Memory w)
                             , mlsEndianness :: !Endianness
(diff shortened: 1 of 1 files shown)


langston-barrett (approved):
If this is legal in ELF, is there any point in warning? Who would want to see the warning, and what would they do about it?

Just an aside, but it would be good to have more clarity on these warnings, generally. Some of them seem to be about features that Macaw doesn't (yet?) support, some of them probably reflect either a malformed ELF file (or equally likely, a malformed understanding of ELF on part of a developer), perhaps there are other categories too (e.g., things like this that are legal but perhaps "suspicious" or unusual)? Some kind of categorization might be helpful for consumers.

RyanGlScott:
I'll keep this open for a while to do some more testing on larger binaries in the wild.

sauclovian-g:
FWIW (apparently I missed this at the time) ordinarily this means your ELF file is messed up, but if you're intentionally messing around, such as the various versions of "smallest possible ELF file" you can find on the internets, you might create overlaps like this on purpose.

PR #489, opened 2025-05-22
Clarify and expand documentation in `Data.Macaw.Memory`

### LINKED ISSUE #488: `Data.Macaw.Memory`: Use "region" or "segment" consistently
This module defines `MemSegment`s, but these are indexed with `RegionIndex`es. This confusion in terminology is widespread, the word "region" and "segment" are used interchangably throughout the Haddocks and even in type and function names. It would be better to decide on what to call this notion once and for all, and use the vocabulary consistently. Renaming functions and types would be a breaking change, but we could presumably provide synonyms and deprecate the old ones first.

langston-barrett:
Fixes #488.


base/src/Data/Macaw/Memory.hs
```diff
@@ -799,22 +841,32 @@ applyRelocsToBytes baseAddr _ pre ioff _ buffer =
...
 --
--- Memory segments with absolute addresses should use a region index
--- of `0`.  Non-zero region indices Absolute segments have a base index of zero, and non-zero values correspond
+-- Non-zero region indices represent addresses that would be chosen at runtime
+-- by the loader (e.g., the virtual address of an ELF segment containing
+-- position-independent code).
+--
+-- The region index 0 indicates an absolute address:
+--
+-- * A 'MemAddr' with a 'addrBase' of 0 is an absolute address.
```
RyanGlScott, line 854:
```suggestion
-- * A 'MemAddr' with an 'addrBase' of 0 is an absolute address.
```

base/src/Data/Macaw/Memory.hs
```diff
@@ -580,6 +618,10 @@ instance Show (Relocation w) where
 -- | A memory chunk describes a contiguous sequence of bytes within a segment.
 --
 -- The parameter denotes the width of a memory address.
+--
+-- Note that the term "region" in this type is not related to the notion of
+-- "region" described in the module-level documentation (i.e., the @Region@
```
RyanGlScott, line 623:
```suggestion
-- Note that the term \"region\" in this type is not related to the notion of
-- \"region\" described in the module-level documentation (i.e., the @Region@
```

base/src/Data/Macaw/Memory.hs
```diff
@@ -1,8 +1,46 @@
 {-|
 Copyright   : (c) Galois Inc, 2015-2018
-Maintainer  : jhendrix@galois.com
-
-Declares 'Memory', a type for representing segmented memory with permissions.
+Maintainer  : Ryan Scott <rscott@galois.com>, Langston Barrett <langston@galois.com>
+
+Declares 'Memory', a type for representing pre-loader memory with permissions.
+This datatype provides an abstraction that is intended to support different
+architectures and executable and object file formats. Crucially, 'Memory' is
```
RyanGlScott, line 7: Lots of "and"s here. Perhaps:

```suggestion
architectures, executable formats, and object file formats. Crucially, 'Memory' is
```

base/src/Data/Macaw/Memory.hs
```diff
@@ -1,8 +1,46 @@
...
+capable of representing relocatable (i.e., position-independent) code and data.
+
+A 'Memory' defines a collection of abstract /regions/ of memory. A region is
+identified by a 'RegionIndex', and represents some an address that would be
```
RyanGlScott, line 11:
```suggestion
identified by a 'RegionIndex' and represents some address that would be
```

base/src/Data/Macaw/Memory.hs
```diff
@@ -1,8 +1,46 @@
...
+chosen at runtime by the loader (e.g., the virtual address of an ELF segment
+containing position-independent code). Thus, in this module, an address
+('MemAddr') consists of a pair of a 'RegionIndex' and an offset into that
+region. A 'MemAddr' with a 'RegionIndex' of 0 represents an /absolute/ address.
+
+'Memory' is essentially a collection of segments. A segment ('MemSegment')
+is a continguous sequence of bytes that will be loaded into runtime memory.
+Segments do not necessarily have a known runtime address. Instead, they use some
+'RegionIndex' as a "base" address, and are located at some fixed offset from
```
RyanGlScott, line 20:
```suggestion
'RegionIndex' as a "base" address and are located at some fixed offset from
```

base/src/Data/Macaw/Memory.hs
```diff
@@ -1,8 +1,46 @@
...
+A @'MemWord' w@ is a @w@-bit machine word. This may be treated as an absolute
+address when @w@ is the width of a pointer ('absoluteAddr').
+
+A 'MemSegmentOff' is notionally a pair @('MemSegment', offset)@, where @offset@
+is an offset into the 'MemSegment'. The 'MemSegmentOff's produced by this module
+are guaranteed to be valid, making it possible to look up the contents of the
+memory they point to.
+
+Each of the above types is parameterized by the number of bits in an address.
+Each type has an alias prefixed with @Arch@ that is instead parameterized by
```
RyanGlScott, line 41: Strictly speaking, I don't know if this is true (e.g., I don't think there's an `Arch` counterpart to `MemWord`). Perhaps we should say that _most_ types have aliases prefixed with `Arch` instead.

base/src/Data/Macaw/Memory.hs
```diff
@@ -1,8 +1,46 @@
...
+
+A 'Memory' defines a collection of abstract /regions/ of memory. A region is
+identified by a 'RegionIndex', and represents some an address that would be
+chosen at runtime by the loader (e.g., the virtual address of an ELF segment
+containing position-independent code). Thus, in this module, an address
+('MemAddr') consists of a pair of a 'RegionIndex' and an offset into that
+region. A 'MemAddr' with a 'RegionIndex' of 0 represents an /absolute/ address.
+
+'Memory' is essentially a collection of segments. A segment ('MemSegment')
+is a continguous sequence of bytes that will be loaded into runtime memory.
```
sauclovian-g, line 18: typo (contiguous)

doc/Design.md
```diff
@@ -147,19 +147,11 @@ Notes:
 
 ## Address Representations
 
-There are three different representations of addresses in Macaw, which often leads to confusion.  This section will attempt to describe them and their intended use cases. The address types are parameterized by the number of bits in the address. Each address type has a type alias prefixed with `Arch` that is instead parameterized by architecture, with the width parameter filled in according to the declared width of the architecture.
+There are four different representations of addresses in Macaw, which often leads to confusion. `MemAddr`, `MemWord`, and `MemSegmentOff` are used in conjunction with the *pre-loader* view of memory. They are disambiguated in the module-level documentation for `Data.Macaw.Memory`.
```
RyanGlScott, line 150: If you want to be pedantic, you could argue that there are actually _five_ such types if you include `MemInt`... perhaps we should change "four" to "several"?

base/src/Data/Macaw/Memory.hs
```diff
@@ -1,8 +1,46 @@
...
+capable of representing relocatable (i.e., position-independent) code and data.
+
+A 'Memory' defines a collection of abstract /regions/ of memory. A region is
+identified by a 'RegionIndex', and represents some an address that would be
+chosen at runtime by the loader (e.g., the virtual address of an ELF segment
+containing position-independent code). Thus, in this module, an address
+('MemAddr') consists of a pair of a 'RegionIndex' and an offset into that
+region. A 'MemAddr' with a 'RegionIndex' of 0 represents an /absolute/ address.
+
+'Memory' is essentially a collection of segments. A segment ('MemSegment')
```
sauclovian-g, line 17: Maybe stick the word "physically" in the first sentence to distinguish from the previous paragraph more?

doc/Design.md
langston-barrett: I would argue that `MemInt` is not a representation of an address, as it is signed. Will change to "several".

base/src/Data/Macaw/Memory.hs
```diff
@@ -580,6 +618,10 @@ instance Show (Relocation w) where
 -- | A memory chunk describes a contiguous sequence of bytes within a segment.
 --
 -- The parameter denotes the width of a memory address.
+--
+-- Note that the term "region" in this type is not related to the notion of
+-- "region" described in the module-level documentation (i.e., the @Region@
+-- in 'RegionIndex').
 data MemChunk (w :: Nat)
```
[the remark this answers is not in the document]
langston-barrett: Agreed, though I'm not sure I feel strongly enough to file an issue :) Feel free to, though!

RyanGlScott (approved):

base/src/Data/Macaw/Memory.hs
sauclovian-g: I guess it seems like the previous paragraph talks about one interpretation of `Memory` and this one talks about a different one (that is, it's a bunch of abstract regions and it's also a bunch of physical segments) and it needs a bit more to contrast the two.

Alternatively I got the wrong idea entirely and the above is off-base, in which case something else should be done :-)

langston-barrett: Yeah, they aren't meant to be two differing interpretations of the same thing. I will think more about how to clarify this. Thanks for highlighting that it is confusing!

sauclovian-g: ok, never mind then. Ping me if you want to talk it over :-)

sauclovian-g (approved):

base/src/Data/Macaw/Memory.hs
langston-barrett: @sauclovian-g Let me know how the new version reads!

base/src/Data/Macaw/Memory.hs
```diff
@@ -1,8 +1,46 @@
...
+that region. A 'MemAddr' with a 'RegionIndex' of 0 represents an /absolute/
+address.
+
+A segment ('MemSegment') is a contiguous sequence of bytes that will be loaded
+into runtime memory. Segments do not necessarily have a known runtime address.
+Instead, they use some 'RegionIndex' as a "base" address and are located at some
+fixed offset from that base. Multiple segments can have the same 'RegionIndex'
+as their base; this indicates that they will have a fixed offset relative to one
+another at runtime. This notion of segment is similar to an ELF segment. It is
+unrelated to the x86 notion of memory segmentation.
```
sauclovian-g, line 24: You might stick an additional explicit note here to the effect that segments with a statically known address live in region 0.

PR #496, opened 2025-05-29
ci: Change OS to self-hosted for Haddock build

langston-barrett:
I think this will fix the doc build...

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -20,9 +20,9 @@ jobs:
     # block, only from `matrix`.
     strategy:
       matrix:
-        os: ["ubuntu-24.04"]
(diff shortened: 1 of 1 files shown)


sauclovian-g (approved):

sauclovian-g:
I wouldn't really expect that to be necessary... note that Ubuntu upstream's been having problems today so it might be sufficient to wait until tomorrow.

langston-barrett:
I retried the build and got the same error:
```
E: Unable to locate package zlibc
E: Package 'python3-distutils' has no installation candidate
E: Package 'python-setuptools' has no installation candidate
```
This is a weird error, though...

sauclovian-g:
It is that. I wonder if some cache needs to be purged...

PR #501, opened 2025-05-30
ci: Switch to GitHub-hosted runners

### LINKED ISSUE #499: ci: Migrate to GitHub-hosted runners
Our CI workflows use self-hosted runners:

https://github.com/GaloisInc/macaw/blob/e0cac7e7f8a8ddc0ce986e17199132aaf2515e60/.github/workflows/ci.yaml#L31

https://github.com/GaloisInc/macaw/blob/e0cac7e7f8a8ddc0ce986e17199132aaf2515e60/.github/workflows/doc.yml#L25

They were introduced in https://github.com/GaloisInc/macaw/commit/a0aaa6e083f34c211ba8049a2a7b0e10fb99c6f4, which says:

> Switch the github actions configuration to use internally-hosted runners, which have enough memory to complete all of the CI jobs

That commit was made in 2021, and there have been [upgrades to the GitHub-provided runners](https://github.blog/news-insights/product-news/github-hosted-runners-double-the-power-for-open-source/) since then, in addition to many GHC improvements. We should try to use GitHub-hosted runners. This would have the following advantages:

1. Maintenance: We wouldn't have to worry about maintaining the runners. Personally, I don't know who is responsible for the current self-hosted runners or how to, say, upgrade the OS on them.

2. Cost: I'm not sure about this, but since Macaw is open-source the GitHub runners might be free.

3. Security: GitHub is likely better than we are at securing infrastructure. (Citation needed 😄)

4. Consistency: All of our other projects use GitHub-provided runners, so we are all familiar with the environment they provide.

langston-barrett:
Fixes #499.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -55,12 +55,12 @@ jobs:
       name: Restore cabal store cache
       with:
         path: |
-          /home/runner/.cabal/store/ghc-${{ matrix.ghc-ver }}
+          ${{ steps.setup-haskell.outputs.cabal-store }}
           dist-newstyle
         # Prefer previous SHA hash if it is still cached
-        key: ${{ env.CACHE_VERSION }}-linux-${{ matrix.ghc-ver }}-${{ hashFiles('cabal.project.freeze') }}-${{ github.sha }}
+        key: ${{ env.CACHE_VERSION }}-${{ matrix.os }}-${{ matrix.ghc-ver }}-${{ hashFiles('cabal.project.freeze') }}-${{ github.sha }}
         # otherwise just use most recent build.
-        restore-keys: ${{ env.CACHE_VERSION }}-linux-${{ matrix.ghc-ver }}-${{ hashFiles('cabal.project.freeze') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -15,20 +15,13 @@ env:
 
 jobs:
   build-linux:
-
-    env:
-      LANG: en_US.UTF-8
-      LC_ALL: en_US.UTF-8
-      TZ: America/Los_Angeles
-      CI_TEST_LEVEL: "2"
```
sauclovian-g, line 23: I think if you want to turn `CI_TEST_LEVEL` down that should happen separately (or at least I guess get called out in the commit message)

...assuming that setting it to 2 actually has any effect, which it kind of looks like it doesn't.

.github/workflows/ci.yaml
```diff
@@ -15,20 +15,13 @@ env:
...
-
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         # Remember to update the GHC version in the documentation workflow as well
         ghc-ver: ["9.4.8", "9.6.6", "9.8.2"]
         cabal: [ '3.10.3.0' ]
-        os: [self-hosted]
+        os: [ubuntu-latest]
```
sauclovian-g, line 24: I suspect you want an actual version here. (otherwise Murphy's Law guarantees that 26.04 appears while we're in the middle of a release cycle or something)

PR #505, opened 2025-06-30
Expose cuRanges

2over12:
Allow consumers to access cuRanges, used downstream in https://github.com/GaloisInc/grease/pull/259


base/src/Data/Macaw/Dwarf.hs
```diff
@@ -22,7 +22,7 @@ module Data.Macaw.Dwarf
     Dwarf.nextCUContext,
     getCompileUnit,
     dwarfCompileUnits,
-    CompileUnit,
+    CompileUnit(cuRanges),
     cuSubprograms,
```
RyanGlScott, line 26: For consistency with how `cuSubprograms` (another field of `CompileUnit`) is exported below, I propose doing the following:

```suggestion
    CompileUnit,
    cuRanges,
    cuSubprograms,
```

This is functionally equivalent to what you have above, except that it makes it more obvious that `CompileUnit` is, for the most part, exported abstractly. (See https://github.com/GaloisInc/macaw/issues/506 for some follow-up work on the "for the most part" bit, which we need not address in this PR.)

RyanGlScott (approved):

PR #509, opened 2025-07-09
Dwarfv5 Implementation

2over12:


base/src/Data/Macaw/Dwarf.hs
```diff
@@ -1414,17 +1422,11 @@ dwarfCompileUnits end sections = do
...
-      let msg = "Could not find " ++ show nm ++ " section."
-      modify $ (:) msg
-      pure $ BS.empty
-    s : r -> do
-      when (not (null r)) $ do
-        let msg = "Found multiple " ++ show nm ++ " sections."
-        modify $ (:) msg
+    [] -> Nothing
+    s : _ -> do
       pure $ Elf.elfSectionData s
```
RyanGlScott, line 1430: I'm a bit alarmed by this change. Before, it would warn if there were multiple ELF sections of the same name, but now it will silently succeed. Is there a good reason not to accumulate warnings here?

2over12: we can't with the API from DWARF since we no longer know what sections are going to be asked for until the parser asks for them. We would need to change the dwarf library to have a warning monad unless we want duplicates to fail (the reason for removing the could not find is it would just fail)

RyanGlScott: I can understand needing to gracefully recover when a section cannot be found (e.g., for sections that are unique to DWARF 5), but what is the rationale for not warning there are duplicate sections?

2over12: We cant do so given the type required by mkSections

2over12: (unless you want to do something really evil related to IO)

2over12: The reason for the type of mksections being Sections -> Maybe is the parser is only a fail monad so you can either: end with failure inside galois-dwarf or keep going, there isnt a monad to warn about things in that parser. That is an artifact of moving section retrieval to a parse time thing depending on version

2over12: Specifically we would need to lift stuff like this https://github.com/GaloisInc/dwarf/blob/82dbd0adbdaf68af924fd918b37ff00c678ac05e/src/Data/Dwarf.hs#L460 into a warning monad of some kind

RyanGlScott (approved):
Excellent, thank you for all your work on this!

PR #515, opened 2025-07-18
Add a MacawProcessAssertion implicit callback for recording macaw memory model predicates

### LINKED ISSUE #429: `macaw-symbolic`: Fine-grained tracking of machine code–specific bad behavior (à la `HasLLVMAnn`)
`macaw-symbolic` is built on top of the `crucible-llvm` memory model, which means that it has the ability to report instances of C-oriented bad behavior (assuming that the underlying machine code adheres to C's memory model conventions). These instances of bad behavior are tracked via the [`HasLLVMAnn`](https://github.com/GaloisInc/crucible/blob/93bfa7f7858eb458f83eb3e0f030bf600ef72365/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Partial.hs#L121) constraint that is threaded throughout `macaw-symbolic`.

In addition to C memory model checks, `macaw-symbolic` also adds a variety of additional assertions that are specific to machine code. These include (but are likely not limited to):

* Checking if reads from and writes to the global address space are within bounds:

  https://github.com/GaloisInc/macaw/blob/186c35da7c35e303ea3e06e026cdf8f0b375ecb5/symbolic/src/Data/Macaw/Symbolic/Memory/Common.hs#L172-L175
  https://github.com/GaloisInc/macaw/blob/186c35da7c35e303ea3e06e026cdf8f0b375ecb5/symbolic/src/Data/Macaw/Symbolic/Memory/Common.hs#L181-L187
* Using a default value when a conditional read's condition does not hold:

  https://github.com/GaloisInc/macaw/blob/186c35da7c35e303ea3e06e026cdf8f0b375ecb5/symbolic/src/Data/Macaw/Symbolic/MemOps.hs#L1095-L1099
* A variety of checks when performing memory operations involving pointers:

  https://github.com/GaloisInc/macaw/blob/186c35da7c35e303ea3e06e026cdf8f0b375ecb5/symbolic/src/Data/Macaw/Symbolic/MemOps.hs#L502-L506
* The `attn` signal (in `macaw-ppc-symbolic`):

  https://github.com/GaloisInc/macaw/blob/186c35da7c35e303ea3e06e026cdf8f0b375ecb5/macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs#L82-L84
* Loss of floating-point precision (in `macaw-ppc-symbolic`):

  https://github.com/GaloisInc/macaw/blob/186c35da7c35e303ea3e06e026cdf8f0b375ecb5/macaw-ppc-symbolic/src/Data/Macaw/PPC/Symbolic/Functions.hs#L235-L246
* Dividing by zero in a division-related instruction (in `macaw-x86-symbolic`):

  https://github.com/GaloisInc/macaw/blob/186c35da7c35e303ea3e06e026cdf8f0b375ecb5/x86_symbolic/src/Data/Macaw/X86/Crucible.hs#L725-L729
* A quotient overflowing in a division-related instruction (in `macaw-x86-symbolic`):

  https://github.com/GaloisInc/macaw/blob/186c35da7c35e303ea3e06e026cdf8f0b375ecb5/x86_symbolic/src/Data/Macaw/X86/Crucible.hs#L767-L769
  https://github.com/GaloisInc/macaw/blob/186c35da7c35e303ea3e06e026cdf8f0b375ecb5/x86_symbolic/src/Data/Macaw/X86/Crucible.hs#L807-L809

Unlike the checks in the `crucible-llvm` memory model, however, these assertions are all performed via `GenericSimError` or `AssertFailureSimError`. As a result, it is not straightforward to catch these assertions and perform subsequent analysis on them after simulation fails.

I propose that `macaw-symbolic` add a constraint similar to `HasLLVMAnn` (perhaps `HasMacawAnn`, for lack of a better name) and use it to track the What4 annotations of each of the terms that give rise to these assertion failures. That way, one can consult the `Map` of bad `macaw-symbolic` behaviors afterwards and match the annotations to the corresponding terms. This would require a fair bit of API churn in order to thread the new `HasMacawAnn` constraint through to the relevant functions, however.

2over12:
Note i removed the fact that this closes the referenced issue as this only adds the error which we would like to refine in Grease currently.

The original issue is #429

### DIFF
--- a/symbolic/src/Data/Macaw/Symbolic/Testing.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/Testing.hs
@@ -461,7 +461,7 @@ simulateAndVerify :: forall arch sym bak ids w solver scope st fs
 simulateAndVerify goalSolver logger bak execFeatures archInfo archVals binfo mmPreset extractor dfi =
   MS.withArchConstraints archVals $ do
     halloc <- CFH.newHandleAllocator
-    let ?processMacawAssert = MSMC.ignoreMacawAssertions
+    let ?processMacawAssert = MSMC.defaultProcessMacawAssertion
     let ?recordLLVMAnnotation = \_ _ _ -> return ()
 
     -- Initialize memory
--- a/symbolic/src/Data/Macaw/Symbolic/Testing.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/Testing.hs
@@ -105,6 +105,7 @@ import qualified What4.ProgramLoc as WPL
 import qualified What4.Protocol.Online as WPO
 import qualified What4.SatResult as WSR
 import qualified What4.Solver as WS
+import qualified Data.Macaw.Symbolic.Memory.Common as MSMC
 
 data TestingException = ELFResolutionError String
   deriving (Show)
@@ -460,6 +461,7 @@ simulateAndVerify :: forall arch sym bak ids w solver scope st fs
 simulateAndVerify goalSolver logger bak execFeatures archInfo archVals binfo mmPreset extractor dfi =
   MS.withArchConstraints archVals $ do
     halloc <- CFH.newHandleAllocator
+    let ?processMacawAssert = MSMC.ignoreMacawAssertions
     let ?recordLLVMAnnotation = \_ _ _ -> return ()
 
     -- Initialize memory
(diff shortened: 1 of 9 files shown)


symbolic/src/Data/Macaw/Symbolic/Testing.hs
```diff
@@ -105,6 +105,7 @@ import qualified What4.ProgramLoc as WPL
 import qualified What4.Protocol.Online as WPO
 import qualified What4.SatResult as WSR
 import qualified What4.Solver as WS
+import qualified Data.Macaw.Symbolic.Memory.Common as MSMC
```
langston-barrett, line 108: This should probably go with the other Macaw imports

PR #516, opened 2025-07-23
Loader for raw binaries

2over12:


macaw-loader/src/Data/Macaw/BinaryLoader.hs
```diff
@@ -6,74 +6,83 @@
...
+whatever loader (e.g. elf), and adding the above constraint
+requires definitions from the macaw architecture specific modules
+(e.g. macaw-ppc), creating a dependency cycle (because macaw-ppc
+should be able to use macaw-loader-ppc). Ergo this constraint must
+be expressed at the usage sites instead.
+-}
+class
+  (MM.MemWidth (MR.ArchAddrWidth arch)) =>
+  BinaryLoader arch binFmt
+  where
```
kquick, line 74: It's hard to tell if this was simply reformatting the declaration and changing the comment markings for the haddock or if there were actual changes applied.  It would be better to make purely format changes in a separate commit (and possibly a separate PR) which would allow easier review of substantive changes and also allow rollback without entanglements.

macaw-loader/src/Data/Macaw/BinaryLoader.hs
```diff
@@ -82,9 +91,10 @@ class (MM.MemWidth (MR.ArchAddrWidth arch)) =>
   -- particular, an ELF file usually specifies an entryPoint address
   -- that is also an address in the symbol table, both of which are in
   -- the data returned by this function.
-  entryPoints :: (X.MonadThrow m) =>
-                 LoadedBinary arch binFmt
-              -> m [MM.ArchSegmentOff arch]
+  entryPoints ::
+    (X.MonadThrow m) =>
+    LoadedBinary arch binFmt ->
+    m [MM.ArchSegmentOff arch]
```
kquick, line 97: (Caveat: I do not generally endorse the use of a linter for common style).  Are these changes being driven by some form of linter or is  there another reason for changing the existing style?

PR #523, opened 2025-08-11
ci: Support GHC 9.10.1, drop GHC 9.4.8

### LINKED ISSUE #463: Support GHC 9.10
https://www.haskell.org/ghc/download_ghc_9_10_1.html

langston-barrett:
Fixes #463.


cabal.project.dist
```diff
@@ -1,3 +1,13 @@
+-- See https://github.com/Happstack/boomerang/issues/12
+allow-newer: boomerang:template-haskell
+-- See https://github.com/GaloisInc/bv-sized-float/issues/8
+allow-newer: bv-sized-float:containers
+-- https://github.com/GaloisInc/grift/issues/13
+allow-newer: grift:containers
+allow-newer: grift:filepath
```
RyanGlScott, line 7: I don't mind bumping the version bounds for `boomerang` and `haggle` for now (as those are outside of the Galois Cinematic Universe), but I wonder if we shouldn't bump the version bumps downstream in `bv-sized-float` and `grift` first before landing this. We control the latter two packages, so it ought to be much more straightforward to handle them.

RyanGlScott (approved):

PR #528, opened 2025-09-10
Add semantics for movmskps

2over12:
Partial work at the moment


langston-barrett:
Closing until we have a moment to revisit.

PR #534, opened 2025-09-18
symbolic: Functions for setting the register state

langston-barrett:


langston-barrett:
This appears to be working in my downstream tests in GREASE: https://github.com/GaloisInc/grease/pull/383

Issue #441, opened 2024-09-23
symbolic: Remove support for stack-spilled arguments


langston-barrett:
`macaw-symbolic` [has some functionality](https://github.com/GaloisInc/macaw/blob/edbcee34946788472469626d46a6deddd429a2f6/symbolic/src/Data/Macaw/Symbolic/Stack.hs#L64) for allocating stack space for spilled arguments. However, many ABIs impose some kind of alignment constraints on the stack pointer. For example, both the AArch32 and x86_64 SysV ABIs specify that the end of the spilled argument list is aligned to `2*w` where `w` is the number of bytes in a word. Since `macaw-symbolic` has no notion of the ABI in use, its ABI-agnostic code could only ever satisfy such alignment constraints on accident. We should just remove it, and clients wishing to spill arguments to the stack should do so with ABI-specific functionality.

PR #535, opened 2025-10-03
aarch32-symbolic: Add `Index` for `pc` (AKA `r15`)

### LINKED ISSUE #440: aarch32-symbolic: `Index` for `pc`
In #438, we introduced `Index`es into `ArchRegStruct` for a lot of the ARM GPRs. It would be nice to have one for `pc`, but it doesn't really have the same shape as the others. It wouldn't be a lot of work to add it, so we should!

langston-barrett:
Fixes #440.


RyanGlScott (approved):
Worth updating `macaw-aarch32-syntax` as well?

RyanGlScott:
Also, the commit message mistakenly refers to #441 instead of #440.

PR #542, opened 2025-10-23
Fix invalid pop semantics in thumb

2over12:
Test case for #543.


RyanGlScott:
Interestingly, `grease` detects an uninitialized stack read on the `test-pushpop.exe` test case, but `macaw-aarch32-symbolic`'s test suite does not. I will need to investigate further to figure out what is causing the discrepancy.

RyanGlScott:
Superseded by https://github.com/GaloisInc/macaw/pull/547.

PR #545, opened 2025-10-27
Expose more regs in syscall ARM

2over12:


langston-barrett:
Closing until we have a moment to revisit.

PR #546, opened 2025-10-31
Add updates for all regs at block entry

2over12:
MacawArchUpdateState does not currently make it possible to track the initial state of registers in blocks coming from block args. This limitation makes it hard to track the state of registers in the crucibleCFG purely by these statements. This PR causes an initial update to be emitted for all registers and is used downstream for https://github.com/GaloisInc/grease/pull/447

I could see an argument against mangling the CFG like this but the regs are already in the incoming reg struct and the CFG is already kinda hard to read as is so im not sure it's worth  creating difficulty downstream for this. (The alternative i suppose would be to capture jumps to blocks in a feature or something and record the register state then)

### DIFF
--- a/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
@@ -1771,7 +1771,7 @@ addRegUpdateForBlock archFns startAddr = do
   void $ evalMacawStmt $ MacawArchStateUpdate (M.segoffAddr startAddr) mp
  where
   updatesFromRegStruct ::
-    (OrdF (M.ArchReg arch)) =>
+    OrdF (M.ArchReg arch) =>
     Assignment (M.ArchReg arch) ctx ->
     CrucGen arch ids s (MapF.MapF (M.ArchReg arch) (MacawCrucibleValue (CR.Atom s)))
   updatesFromRegStruct = foldlMFC' (\mp reg -> do


2over12:
We can also do this same thing after calls to handle updating that missed state if we find that acceptable as well. 


Hmm i dont think this should be required actually looking at it? ParsedCall should result in a fresh jump to the return addr anyways which should have its own UpdateState

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -1755,6 +1756,22 @@ addSwitch blockLabelMap idx possibleAddrs = do
...
+ where
+  updatesFromRegStruct ::
+    (OrdF (M.ArchReg arch)) =>
+    Assignment (M.ArchReg arch) ctx ->
+    CrucGen arch ids s (MapF.MapF (M.ArchReg arch) (MacawCrucibleValue (CR.Atom s)))
+  updatesFromRegStruct Empty = pure MapF.empty
+  updatesFromRegStruct (rst :> reg) = do
+    rstOf <- updatesFromRegStruct rst
+    rval <- getRegValue reg
+    pure $ MapF.insert reg (MacawCrucibleValue rval) rstOf
```
langston-barrett: Looks like perhaps this could be written as `foldlMFC` or similar? Maybe not though, and no biggie either way.

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -1755,6 +1756,22 @@ addSwitch blockLabelMap idx possibleAddrs = do
   mapM_ addStmt stmts
   addTermStmt termStmt
 
+
+addRegUpdateForBlock :: (OrdF (M.ArchReg arch)) => MacawSymbolicArchFunctions arch -> M.ArchSegmentOff arch -> CrucGen arch ids s ()
```
langston-barrett: I know that this is a non-exported function but might be worth a Haddock explaining *why* we want to do this.

langston-barrett (approved):
Seems reasonable.

> Hmm i dont think this should be required actually looking at it? ParsedCall should result in a fresh jump to the return addr anyways which should have its own UpdateState

Does this comment also apply to calls to Haskell overrides, like for libc functions? Syscalls? Might be worth documenting known limitations, perhaps in the Haddock for `MacawArchStateUpdate`. In fact, that Haddock likely needs an update anyway if nothing else just to say that there is one at the beginning of each block.

langston-barrett:
I personally am not concerned about "clutter" in the CFGs, as you note, they are already quite unwieldy.

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -1755,6 +1756,22 @@ addSwitch blockLabelMap idx possibleAddrs = do
...
+addRegUpdateForBlock archFns startAddr = do
+  mp <- updatesFromRegStruct (crucGenRegAssignment archFns)
+  void $ evalMacawStmt $ MacawArchStateUpdate (M.segoffAddr startAddr) mp
+ where
+  updatesFromRegStruct ::
+    (OrdF (M.ArchReg arch)) =>
```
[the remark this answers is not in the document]
2over12: ah hmm the default configuration of fourmolu without say GREASE's config is to add parens :( I guess I should probably disable format changed lines

Issue #266, opened 2022-03-07
Simulation failure on AArch32: `ldm` semantics perform an unnecessary memory read


RyanGlScott:
Consider this C program, appropriately decorated for use in `macaw`'s test suite:

```c
// bug.c
int __attribute__((noinline)) test_string_lit() {
  char x[5] = "ABCD";
  return x[0] == 'A';
}

void _start() {
  test_string_lit();
}
```

Compile this program with the an AArch32 version of `musl-gcc`:

```
$ armv7m-linux-musleabi-gcc -nostdlib -static -fno-stack-protector -no-pie bug.c -o bug.exe
$ armv7m-linux-musleabi-objdump -d bug.exe 

bug.exe:     file format elf32-littlearm


Disassembly of section .text:

00010074 <test_string_lit>:
   10074:       e52db004        push    {fp}            ; (str fp, [sp, #-4]!)
   10078:       e28db000        add     fp, sp, #0
   1007c:       e24dd00c        sub     sp, sp, #12
   10080:       e59f2038        ldr     r2, [pc, #56]   ; 100c0 <test_string_lit+0x4c>
   10084:       e08f2002        add     r2, pc, r2
   10088:       e24b300c        sub     r3, fp, #12
   1008c:       e8920003        ldm     r2, {r0, r1}
   10090:       e5830000        str     r0, [r3]
   10094:       e2833004        add     r3, r3, #4
   10098:       e5c31000        strb    r1, [r3]
   1009c:       e55b300c        ldrb    r3, [fp, #-12]
   100a0:       e3530041        cmp     r3, #65 ; 0x41
   100a4:       03a03001        moveq   r3, #1
   100a8:       13a03000        movne   r3, #0
   100ac:       e20330ff        and     r3, r3, #255    ; 0xff
   100b0:       e1a00003        mov     r0, r3
   100b4:       e28bd000        add     sp, fp, #0
   100b8:       e49db004        pop     {fp}            ; (ldr fp, [sp], #4)
   100bc:       e12fff1e        bx      lr
   100c0:       0000004c        .word   0x0000004c

000100c4 <_start>:
   100c4:       e92d4800        push    {fp, lr}
   100c8:       e28db004        add     fp, sp, #4
   100cc:       ebffffe8        bl      10074 <test_string_lit>
   100d0:       e1a00000        nop                     ; (mov r0, r0)
   100d4:       e8bd8800        pop     {fp, pc}
```

[...]

Issue #543, opened 2025-10-23
POP_T1 invalid semantics


2over12:
On thumb pop semantics result in a macaw CFG with a read at the original stack pointer:
```
function test_pop @ 0x10099
0x10099:
  ;  _R13 = stack_frame + 0
  # 0x10099 0x10098: PUSH_T1(xxxxxxxx.1011010x) M 0, register_list 128
  # 0x10098: PUSH_T1(xxxxxxxx.1011010x) M 0, register_list 128
  r67 := (bv_add _R13_0 (0xfffffffc :: [32]))
  write_mem r67 _R7_0
  # 0x10098: {PSTATE_IT => 0x0 :: [8]
             ;_PC => 0x1009b
             ;_R13 => r67
             ;__AssertionFailure => false
             ;__UndefinedBehavior => false
             ;__UnpredictableBehavior => false
             ;() => ()_0}
  # 0x1009b 0x1009a: POP_T1(xxxxxxxx.1011110x) P 0, register_list 128
  # 0x1009a: POP_T1(xxxxxxxx.1011110x) P 0, register_list 128
  r79 := read_mem _R13_0 (bvle4)
  r87 := read_mem r67 (bvle4)
  # 0x1009a: {PSTATE_IT => 0x0 :: [8]
             ;PSTATE_T => 0x1 :: [1]
             ;_PC => 0x1009c
             ;_R13 => _R13_0
             ;_R7 => r87
             ;__AssertionFailure => false
             ;__BranchTaken => false
             ;__UndefinedBehavior => false
             ;__UnpredictableBehavior => false
             ;() => ()_0}
```

This read is invalid because only the read at r67 has been initialized. downstream issue [here](https://github.com/GaloisInc/grease/issues/432), test case here: https://github.com/GaloisInc/macaw/pull/542

RyanGlScott:
I suspect that this and https://github.com/GaloisInc/macaw/issues/266 share a root cause.

RyanGlScott:
This should be fixed as of https://github.com/GaloisInc/macaw/pull/547. It isn't currently possible to test for this in `macaw-aarch32-symbolic`'s test suite due to due limitations in `Data.Macaw.Symbolic.Testing`'s stack modeling, however (see https://github.com/GaloisInc/macaw/issues/430). As such, I am going to add a test case to `grease` instead, which is equipped to test for this.

PR #547, opened 2025-11-01
Fix redundant mem reads by embedding path conditions

danmatichuk:
Fixes: https://github.com/GaloisInc/macaw/issues/266

Translates "uf_cond_read_mem" calls into "condReadMem", which takes the path condition as an argument, and emits an undefined value if the condition is false.


RyanGlScott:
This looks great! I think the only thing that's missing is to add a `macaw-aarch32-symbolic` test case that is inspired by the program from https://github.com/GaloisInc/macaw/issues/266. Since this critically relies on the use of an `ldm` instruction, it might be wise to check in a test case as an assembly file so that we can guarantee that the `ldm` instruction will appear in the binary. Something like this ought to suffice:

```
.global test_ldm
.syntax unified
.type	test_ldm, %function
test_ldm:
  push    {fp}
  add     fp, sp, #0
  sub     sp, sp, #12
  ldr     r2, =x
  sub     r3, fp, #12
  ldm     r2, {r0, r1}
  str     r0, [r3]
  add     r3, r3, #4
  strb    r1, [r3]
  ldrb    r3, [fp, #-12]
  cmp     r3, #65
  moveq   r3, #1
  movne   r3, #0
  and     r3, r3, #255
  mov     r0, r3
  add     sp, fp, #0
  pop     {fp}
  bx      lr

.global _start
.syntax unified
_start:
  push    {fp, lr}
  add     fp, sp, #4
  bl      test_ldm
  nop     {0}
  pop     {fp, pc}

.data
x:
  .asciz "ABCD"
```

I have also confirmed that this fixes the issue in https://github.com/GaloisInc/macaw/issues/543 by running the test case from https://github.com/GaloisInc/macaw/pull/542 in `grease`. There isn't a good way to add this as a `macaw-aarch32-symbolic` test case, since it's [not equipped to detect uninitialized stack reads](https://github.com/GaloisInc/macaw/issues/430) like `grease` is. As such, I think we should just close https://github.com/GaloisInc/macaw/issues/543 and https://github.com/GaloisInc/macaw/pull/542 after this PR lands. I can take care of adding a test case to `grease` afterwards.

danmatichuk:
@RyanGlScott Can I see the source for that test? I naively compiled it and ran it through the macaw-aarch32-symbolic test harness and it failed to verify.

danmatichuk:
> Huh, you're right—that _doesn't_ verify in the `macaw-aarch32-symbolic` test suite. Odd, that—it behaves as expected when I run it through `grease`. I'll need to investigate a bit more to figure out why we're getting different results in `macaw-aarch32-symbolic`'s test suite.
> 
> On a slightly related note, I looked at the `macaw-aarch32-dump discover` output of this binary, and I noticed something strange. An easy way to reproduce the issue with a binary that is already checked into version control is to do the following:
> 
> ```
> $ cabal -v0 run exe:macaw-aarch32-dump -- discover macaw-aarch32-symbolic/tests/pass/identity.unopt.exe
> function test_identity @ 0x100b8
> <snip>
>   # 0x100d4 0x100d4: LDR_i_A1_post(xxxxxxxx.xxxxxxxx.x001xxxx.xxxx0100) P 0, Rn 13, Rt 11, U 1, W 0, cond 14, imm12 4
>   # 0x100d4: LDR_i_A1_post(xxxxxxxx.xxxxxxxx.x001xxxx.xxxx0100) P 0, Rn 13, Rt 11, U 1, W 0, cond 14, imm12 4
>   r339 := read_mem r186 (bvle4)
>   r349 := read_mem r186 (bvle4)
>   # 0x100d4: {PSTATE_T => 0x0 :: [1]
>              ;_PC => 0x100d8
>              ;_R11 => r349
>              ;_R13 => _R13_0
>              ;__AssertionFailure => false
>              ;__BranchTaken => false
>              ;__UndefinedBehavior => false
>              ;__UnpredictableBehavior => false
>              ;() => ()_0}
> <snip>
> ```
> 
> Somehow, a single `ldr` instruction is translated to multiple `read_mem r186 (bvle4)` calls, and only one of the calls is actually used! This didn't happen prior to these changes. Do you know what is going on here?

The multiple reads are almost certainly a result of the conditional reads affecting term sharing in the underlying What4 representation, meaning that some lazily evaluated code might get executed twice when it didn't before. For example if you have "If A then read(X) else (read(X) + 1)" then originally that "read(X)" would get turned into a single monadic action. Now it looks like: "if A then condRead(X,A) else condRead(X, not A) + 1", which can result in multiple reads during discovery if A is symbolic since both branches need to execute. I opted not to actually emit conditional reads since there's not really much value in it (and it probably would break other things), but if I did I think you'd see some condition on one read, and its negation on another.

In general the translation doesn't work very hard to de-duplicate reads, as it doesn't track them the same way it does writes (if it did, likely we wouldn't have had the original issue at all).

danmatichuk:
I was thinking at some point that we could opt to only do the conditional reads pass only on instructions we've identified as problematic, which would leave LDR unchanged. We can keep this option in our back pocket if the redundant reads this introduces actually become an issue.

PR #568, opened 2026-03-20
symbolic: Enable assertions in CI

### LINKED ISSUE #567: Enable assertions in CI
We have places in Macaw that use `Control.Exception.assert`:

https://github.com/GaloisInc/macaw/blob/144eb3ed7b85bea8c80855b77fa691696023be64/x86/src/Data/Macaw/X86/Monad.hs#L912

However, Cabal compiles at `-O1` by default, which disables such assertions. We should do the same thing as https://github.com/GaloisInc/crucible/pull/1559 to enable them in CI (and test that they are enabled).

langston-barrett:
Fixes #567.


langston-barrett:
This only enables assertions for macaw-symbolic, not for other packages that have them like macaw-x86. Will leave the issue open for doing the same for those other packages.

RyanGlScott (approved):

PR #582, opened 2026-03-30
Tests for strict/lazy memory model consistency properties

langston-barrett:


langston-barrett:
Too complex for tests.

PR #594, opened 2026-04-02
x86-symbolic: Add concrete CRC32 test

### LINKED ISSUE #577: Strict memory model: Off by one?
Feels like this should be `CO`:

https://github.com/GaloisInc/macaw/blob/0d34f687ec8a1a9401e42ff255938ccabfa8ec03/symbolic/src/Data/Macaw/Symbolic/Memory.hs#L440

langston-barrett:
Add a test that hammers concrete reads from concrete memory. Required switching the tests to use an online backend.

Also, fixes #577.


x86_symbolic/tests/pass/crc32-global-input.c
```diff
@@ -0,0 +1,127 @@
+// Test that symbolic execution can efficiently handle CRC-32 over a static
+// const input array. Both the CRC table and the input data live in .rodata,
+// so every memory read in the loop hits the concreteUnmutatedGlobalRead fast
+// path. With the optimization the solver can determine the checksum is the
+// concrete value 0x68d892cf and prove the assertion without path explosion.
+
+static const unsigned long crc_table[256] = {
```
RyanGlScott: My compiler knowledge is lacking here: is it actually guaranteed that a C compiler will place this in the `.rodata` section specifically? It might be worth giving this (and `input_data`) an `__attribute__((section(".rodata")))` attribute just to be on the safe side.

macaw-aarch32-symbolic/tests/Main.hs
```diff
@@ -320,7 +322,5 @@ toSavedSMTSessionPath adapter funName p = "/tmp" </> filename <.> "smtlib2"
                       , "_"
                       , WS.solver_adapter_name adapter
                       ]
-
-escapeSlash :: Char -> Char
-escapeSlash '/' = '_'
-escapeSlash c = c
+    escapeSlash '/' = '_'
+    escapeSlash c = c
```
RyanGlScott: What is the purpose of this change here (and in other symbolic `test/Main.hs` files)? Previously, `escapeSlash` was only defined once, but now it's defined twice within the same file as separate local functions to `toSavedMacawPath` and `toSavedSMTSessionPath`.

symbolic/src/Data/Macaw/Symbolic/Memory.hs
```diff
@@ -437,7 +437,7 @@ populateSegmentChunk _ bak mmc mem symArray seg addr bytes ptrtable = do
   let ?ptrWidth = MC.memWidth mem
   let abs_addr = toAbsoluteAddr addr
   let size = length bytes
-  let interval = IM.IntervalOC abs_addr (abs_addr + fromIntegral size)
+  let interval = IM.IntervalCO abs_addr (abs_addr + fromIntegral size)
```
RyanGlScott: This change feels rather unrelated to the use of online solvers, so I'd propose splitting it out into its own commit (where the commit message says "fixes #577").

RyanGlScott (approved):

x86_symbolic/tests/pass/crc32-global-input.c
langston-barrett: done!

macaw-aarch32-symbolic/tests/Main.hs
langston-barrett: Agent slop, my apologies for not catching it :/

symbolic/src/Data/Macaw/Symbolic/Memory.hs
langston-barrett: I thought tthat this was necessary but it turns out not to be. Removed until we have a test-case that demonstrates the fix.

PR #596, opened 2026-04-13
base, x86: Drop `lens` for `microlens-*`

langston-barrett:
All of their dependencies now use microlens. This makes build graphs "wider" and less "deep", exposing more opportunities for parallelism. For uses of macaw without the *-symbolic packages, this can make build
graphs substantially smaller.


base/src/Data/Macaw/CFG/Core.hs
```diff
@@ -613,9 +619,9 @@ getBoundValue r (RegState m) =
 -- Without this rule, boundValue gets left as a higher-order function,
 -- making its uses VERY slow.
 {-# RULES
-      "boundValue/get" forall rs r. get (boundValue r) rs = getBoundValue r rs
+      "boundValue/view" forall rs r. view (boundValue r) rs = getBoundValue r rs
   #-}
--- Note that this rule seems to cover (^.) as well as get, which is
+-- Note that this rule seems to cover (^.) as well as view, which is
 -- fortunate since a parsing bug makes it impossible to mention (^.)
 -- in a rule.
```
langston-barrett, line 626: I really have no way of knowing whether this update breaks stuff. Presumably it would for API consumers using lens, which would have to switch to `microlens` for this to fire. I don't even know if it does fire any more. I can do some profiling, e.g., using `macaw-dump` if that would be helpful.

Then again... for consumers who see a performance issue here... maybe try using the less polymorphic alternative directly, rather than relying on what seems like pretty fragile `RULES` behavior? I could even see an argument for removing this at this point.

RyanGlScott: I'm personally not very worried about this for the reasons you mention.

RyanGlScott (approved):

PR #605, opened 2026-05-05
Fix bug in jump table simulation

langston-barrett:
See https://github.com/GaloisInc/grease/issues/677 and the commit message.


symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -1556,8 +1556,9 @@ addMacawParsedTermStmt blockLabelMap externalResolutions thisAddr tstmt = do
       let flbl = parsedBlockLabel blockLabelMap falseAddr
       addTermStmt $! CR.Br crucCond tlbl flbl
     M.ParsedLookupTable _layout regs idx possibleAddrs -> do
+      idxVal <- v2c idx
       setMachineRegs =<< createRegStruct regs
```
RyanGlScott: Am I understanding correctly that it's important that `v2c idx` be called _before_ calling `setMachineRegs =<< createRegStruct regs`?

Interestingly, there is now a bit of incongruity between the `addSwitch` and `addIPSwitch` functions. Previously, they both took a `Value` as an argument, and they both called `v2c` within the bodies of the functions. Now, only `addIPSwitch` does this. I think this is the right order for `addIPSwitch`, since the comments [here](https://github.com/GaloisInc/macaw/blob/a7b3f31e32054f2f947fb4443373ffa2520d4c29/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs#L1622-L1624) suggest that the IP value should be taken from the register state at the terminator (i.e., _after_ calling `setMachineRegs =<< createRegStruct regs` [here](https://github.com/GaloisInc/macaw/blob/a7b3f31e32054f2f947fb4443373ffa2520d4c29/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs#L1592)).

To make this a little clearer, I wonder if we should do the following:

* Pass both the `Value` (`idx` for `addSwitch` and `macaw_ip` for `addIPSwitch`) and the registers as explicit arguments to `addSwitch`/`addIPSwitch`.
* Within the bodies of `addSwitch`/`addIPSwitch`, call `v2c` and `setMachineRegs =<< createRegStruct regs` in the correct orders, making a note of why the order is important in each function.

Does that sound reasonable?

langston-barrett: Done!

RyanGlScott (approved):

PR #608, opened 2026-05-06
perf(symbolic): read register struct less in `addRegUpdateForBlock`

langston-barrett:
`addRegUpdateForBlock` is responsible for producing `MacawArchStateUpdate` metadata statements in generated Crucible CFGs. Before this change, it called `getRegValue` for each architectural register. Each call emitted a redundant `ReadReg` atom to read the entire register struct from the mutable Crucible register. For x86-64 (~50 registers), this produced many duplicate `ReadReg` statements per basic block in the generated Crucible CFG.

Instead, read the struct once up front and extract each field from the cached atom. On a ~9MB x86 ELF binary (2665 functions, 70648 macaw blocks, 77828 crucible blocks), this reduces heap allocations by ~9.6% (573.8 GB -> 518.5 GB) and total runtime by ~7.5% (331.8s -> 306.9s). The generated Crucible statement counts are unchanged.

### DIFF
--- a/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
@@ -1767,16 +1767,28 @@ at entry to the next block.
 -}
 addRegUpdateForBlock :: OrdF (M.ArchReg arch) => MacawSymbolicArchFunctions arch -> M.ArchSegmentOff arch -> CrucGen arch ids s ()
 addRegUpdateForBlock archFns startAddr = do
-  mp <- updatesFromRegStruct (crucGenRegAssignment archFns)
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Good catch!

I wonder if we should rethink the API for `getRegValue`, since it makes it all too easy to be wasteful if you call it repeatedly with the same register struct. For instance, I found another part of the code that could likely be made faster by factoring out all of the register struct reads into a single place:

https://github.com/GaloisInc/macaw/blob/c98f3a142001e79d85c905e3c03db9e6063bd423/x86/src/Data/Macaw/X86/Generator.hs#L359

There might be other instances of this as well---I haven't checked.

langston-barrett:
#612 is better.

PR #612, opened 2026-05-06
perf(symbolic): New start-of-block extension statement

langston-barrett:
8b7f419 added `addRegUpdateForBlock`, which is responsible for producing `MacawArchStateUpdate` metadata statements at the beginning of a block. It did this using N `getRegValue` calls, which turns into N `ReadReg` statements, followed by N `GetStruct`s (for x86_64, N ~= 70 registers).

As it turns out, this is unnecessary: we can simply embed the entire `ArchRegStruct` value directly via a new `MacawBlockRegState` statement. This results in just one `ReadReg` statement per block. It's still necessary to embed this as a statement, so that consumers like GREASE can track register updates entirely via an extension evaluator.

We need to keep `MacawArchStateUpdate` for intra-block cases where not all registers are touched (i.e., instructions).

On a representative binary:

|                      |     baseline |   MacawBlockRegState |   delta |
| -------------------- | -----------: | -------------------: | ------: |
| Crucible statements  |    8,060,303 |            3,158,656 | −61%    |
| Bytes allocated      |       574 GB |               335 GB | −42%    |
| Bytes copied (GC)    |      38.1 GB |              17.8 GB | −53%    |
| Max residency        |     1,534 MB |             1,523 MB | ~0%     |
| Total memory in use  |     4,470 MB |             4,439 MB | ~0%     |
| MUT time             |      357.0 s |              216.3 s | −39%    |
| GC time              |       30.1 s |               13.5 s | −55%    |
| **Total wall time**  |    **387 s** |            **229 s** | **−41%** |
| Gen-0 collections    |      137,673 |               80,226 | −42%    |
| Gen-1 collections    |           20 |                   15 | −25%    |

### DIFF
--- a/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
@@ -370,6 +370,11 @@ data MacawStmtExtension (arch :: K.Type)
                         -> !Text.Text
                         -> MacawStmtExtension arch f C.UnitType
 
+  -- | Metadata recording the full register state at the start of a block.
+  MacawBlockRegState :: !(M.ArchMemAddr arch)
+                     -> !(f (ArchRegStruct arch))
+                     -> MacawStmtExtension arch f C.UnitType
+
   -- NOTE: The Ptr* operations below are statements and not expressions
   -- because they need to read the memory variable, to determine if their
   -- inputs are valid pointers.
@@ -636,6 +641,8 @@ instance (C.PrettyApp (MacawArchStmtExtension arch),
         in sexpr "macawArchStateUpdate" [pretty addr, encloseSep lbrace rbrace semi (MapF.foldrWithKey prettyArchStateBinding [] m)]
       MacawInstructionStart baddr ioff t ->
         sexpr "macawInstructionStart" [ pretty baddr, pretty ioff, viaShow t ]
+      MacawBlockRegState addr regs ->
+        sexpr "macawBlockRegState" [ pretty addr, f regs ]
       PtrEq _ x y    -> sexpr "ptr_eq" [ f x, f y ]
       PtrLt _ x y    -> sexpr "ptr_lt" [ f x, f y ]
       PtrLeq _ x y   -> sexpr "ptr_leq" [ f x, f y ]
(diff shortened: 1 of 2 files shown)


symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -370,6 +370,11 @@ data MacawStmtExtension (arch :: K.Type)
                         -> !Text.Text
                         -> MacawStmtExtension arch f C.UnitType
 
+  -- | Metadata recording the full register state at the start of a block.
```
RyanGlScott, line 373: Should users interpret `MacawBlockRegState` as "read-only"? That is, this simply displays a view of the current state of the registers at the start of a block, rather than marking the presence of register-updating instructions à la `MacawArchStateUpdate`?

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -676,6 +684,132 @@ instance C.TypeApp (MacawArchStmtExtension arch)
...
+-- address). Memory accesses, fresh symbolic values, function-handle lookups,
+-- and metadata statements must stay impure.
+instance C.IsPureStmt (MacawStmtExtension arch) where
+  pureStmtCompareFC cmp x y = case (x, y) of
+    (MacawGlobalPtr w1 a1, MacawGlobalPtr w2 a2) ->
+      case compareF w1 w2 of
+        LTF -> Just LTF
+        GTF -> Just GTF
+        EQF -> Just (fromOrdering (compare a1 a2))
+    (PtrEq w1 x1 y1, PtrEq w2 x2 y2) ->
```
RyanGlScott: A stylistic thing: I usually like to enumerate the constructors in a comparison-style function like this:

```hs
  pureStmtCompareFC cmp x y = case (x, y) of
    -- Equal cases
    (MacawGlobalPtr w1 a1, MacawGlobalPtr w2 a2) ->
      ...
    (PtrEq w1 x1 y1, PtrEq w2 x2 y2) ->
       ...
    -- Other cases go here...

    -- Unequal cases
    (MacawGlobalPtr {}, _) ->
       Nothing
    (PtrEq {}, _) ->
       Nothing
    -- Other cases go here...
```

That way, GHC will warn us if we ever add new constructors, which will remind us to update this accordingly. This requires some more cases, but the number of cases is linear rather than quadratic.

RyanGlScott (approved):

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
langston-barrett: Oh whoops, this part was an unintentional commit from an unrelated change I was benchmarking!

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
langston-barrett: Yes!

PR #618, opened 2026-05-15
base: Add tests for abstract domains

langston-barrett:
Found a bug in `joinAbsValue`!


base/src/Data/Macaw/AbsDomain/AbsState/Internal.hs
```diff
@@ -0,0 +1,1516 @@
...
+    | StridedInterval si <- v,  FinSet s <- v' =
+      let si_new = SI.fromFoldable (SI.typ si) s
+      in if si_new `SI.isSubsetOf` si
+           then return Nothing
+           else return $ go si si_new
+    | StridedInterval si <- v,  CodePointers s b <- v' = do
+      addWords s
+      let (wordSet, _) = partitionAbsoluteAddrs s b
+          si_new = SI.fromFoldable (SI.typ si) wordSet
+      return $ if si_new `SI.isSubsetOf` si then Nothing else go si si_new
```
langston-barrett, line 479: This was the only logic that changed in this file - this function is supposed to return `Nothing` if the new value is not actually a refinement of the old.

RyanGlScott: If feasible, it would be worth splitting this change out into its own commit. Otherwise, the diff makes it appear as though this is just moving code around.

RyanGlScott (approved):

PR #619, opened 2026-05-15
symbolic: Encode abstract values into Crucible CFGs

### LINKED ISSUE #614: symbolic: Encode abstract values into Crucible CFGs
## The problem

Consider the following C program:
```c
// jump-table.c
int needs_escape(unsigned char c)
{
    switch (c) {
    case 0x00 ... 0x1f:
    case 0x7f ... 0xff:
    case '\\': case '%': case '#': case '/':
    case '(': case ')': case '<': case '>':
    case '[': case ']': case '{': case '}':
        return 1;
    default:
        return 0;
    }
}

int main(int argc, char* argv[]) {
    return needs_escape(argc);
}
```
Let's look at the Macaw CFG for `needs_escape`, in particular the block with the jump table:
```sh
clang-14 -O2 -g -fPIC -o jump-table.elf jump-table.c
cabal run macaw-x86-dump -- discover jump-table.elf
```
```
segment1+0x1144:
  ;  rsp = stack_frame + 0
  ;  rcx in (64 0 90)
  ;  rdi in (64 35 125)
  ;  x87top in (3 7 7)
  # segment1+0x1144 movzx  ecx,cl
  r67 := (trunc rcx_0 8)
--snip --
  # segment1+0x1155 jmp    rcx
  # {1}+0x1155: {rip => r79}
  ijump r69
    0 :-> segment1+0x113b
    1 :-> segment1+0x1157
```
Take note of those abstract domain values for `rcx` and `rdi` in the comments.

Now, let's look at the Crucible CFG:
```sh
cabal run macaw-x86-dump -- discover --crucible jump-table.elf
```
The jump table is encoded as a long chain of if-then-elses, eventually terminating in an error case:

<details>

<summary>Crucible CFG</summary>

```
%4
  % jump-table.elf: segment1+0x1155
  $2 = bVLit(64, BV 1)
  % jump-table.elf: segment1+0x1155
  $3 = extensionApp((bits_to_ptr_64 $2))
  % jump-table.elf: segment1+0x1155
  $4 = (ptr_eq $1 $3)
  % jump-table.elf: segment1+0x1155
  br $4 %96($0) %5($0, $1)
  % no postdom
%5
  % jump-table.elf: segment1+0x1155
  $2 = bVLit(64, BV 2)
  % jump-table.elf: segment1+0x1155
  $3 = extensionApp((bits_to_ptr_64 $2))
  % jump-table.elf: segment1+0x1155
  $4 = (ptr_eq $1 $3)
  % jump-table.elf: segment1+0x1155
  br $4 %2($0) %6($0, $1)
  % no postdom
-- snip --
%94
  % jump-table.elf: segment1+0x1155
  $0 = stringLit("Index not in lookup table")
  % jump-table.elf: segment1+0x1155
  error $0
  % no postdom
```
  
</details>

Note that **the abstract values don't appear in the Crucible CFG.** You might ask: why does this matter? Crucible does symbolic execution, so the path condition will encode *strictly more* information than an all-paths abstract interpretation. And that's true! Crucible has all of the information it needs to do a sound analysis.

The problem is one of performance, and it's an interesting interaction with Macaw's lazy memory model. The point of the lazy memory model is to avoid loading the entire binary into an SMT array if the program has only read from or written to a small portion of its address space. In practice, this works by extracting the abstract value from the What4 expression representing the program:

https://github.com/GaloisInc/macaw/blob/e1e0ae8a9e3e7ab7a21aadb23dd3cdba9ef1e8a4/symbolic/src/Data/Macaw/Symbolic/Memory/Lazy/Internal.hs#L908-L920

And then loading only those segments of memory that may overlap with it. All of this is described in more detail here:

https://github.com/GaloisInc/macaw/blob/e1e0ae8a9e3e7ab7a21aadb23dd3cdba9ef1e8a4/symbolic/src/Data/Macaw/Symbolic/Memory/Lazy/Internal.hs#L980

Note that the What4 abstract value is *entirely unrelated* to the one seen in the Macaw CFG. It is not generated from an all-paths abstract interpretation, but rather bottom up from the symbolic bitvector expression (e.g., `ite b 5 7 + 3` would have the range `[8, 10]` as its What4 abstract value). In particular, its *precision* is independent of the Macaw abstract value--it could be more, or less precise.

Consider the case where it is less precise, which will be true in the above program. The path condition and the Macaw abstract value both encode the constraint that the jump table index is in the narrow range `[0, 90]`. However, the What4 expression turns out to have a wholly unconstrained abstract value `[0, 2^64)`. When macaw-symbolic/Crucible executes the jump at runtime, it will thus load *the entire program* into the SMT array. This is catastrophic for large binaries, and analysis immediately halts/dies/spins.

## The solution

So what can we do? Luckily, we *know* Macaw *already has* tight bounds on this pointer from its abstract interpretation - otherwise, it wouldn't have been able to recover the jump table! Let's just use them.

I propose we make a new Crucible Macaw expression extension `SetAbstractValue :: BV -> Bounds -> BV`. This would be emitted at the beginning of each basic block for each register for which Macaw has computed nontrivial bounds. At symbolic execution time, the Macaw expression extension would check whether or not the bound computed by Macaw is more precise than the current What4 bound, and if so, use [`unsafeSetAbstractValue`](https://hackage-content.haskell.org/package/what4-1.7.3/docs/What4-Interface.html#v:unsafeSetAbstractValue). We may or may not want to use [`annotateTerm`](https://hackage-content.haskell.org/package/what4-1.7.3/docs/What4-Interface.html#v:annotateTerm) at the same time, too (see the Haddocks on both for why we might want to do that).

langston-barrett:
Modest CFG expansion on the GREASE test suite:

- x86-64: ~2.3%
- ppc32: ~0.2%
- armv7l: ~0.2%

The meet of both analyses really does fire and help.

Fixes #614.

### DIFF
--- a/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
@@ -1829,29 +1829,34 @@ absValueToBVRange ::
   M.MemWidth w =>
   NatRepr n ->
   MA.AbsValue w (M.BVType n) ->
-  Maybe (Integer, Integer)
+  Maybe (BV.BV n, BV.BV n)
 absValueToBVRange w av =
   case av of
     MA.FinSet s | not (Set.null s) ->
-      checkRange (Set.findMin s, Set.findMax s)
+      checkRange (Set.findMin s) (Set.findMax s)
     MA.StridedInterval si | MASI.size si > 0 ->
-      checkRange (MASI.base si, MASI.intervalEnd si)
+      checkRange (MASI.base si) (MASI.intervalEnd si)
     MA.SubValue (n' :: NatRepr n') innerAv
       | Just (lo, hi) <- absValueToBVRange n' innerAv ->
-          checkRange (liftInnerRange w n' lo hi)
+          let (loLifted, hiLifted) = liftInnerRange w n' lo hi
+          in Just (loLifted, hiLifted)
     _ -> Nothing
   where
-    checkRange r@(lo, hi)
-      | lo < 0 || hi > maxBnd || lo > hi =
-          MSP.panic
-            MSP.Symbolic
-            "absValueToBVRange"
-            [ "AbsValue bounds outside declared bitvector width"
-            , "width: " ++ show w
-            , "bounds: " ++ show r
-            ]
-      | otherwise = Just r
-    maxBnd = 2 ^ natValue w - 1
+    checkRange lo hi =
+      let loBV = BV.mkBV w lo
+          hiBV = BV.mkBV w hi
+          -- Check: lo < 0 (must check as Integer before constructing BV)
+          -- Check: hi > maxUnsigned w becomes: maxUnsigned w < hi
+          -- Check: lo > hi becomes: hi < lo
+      in if lo < 0 || BV.maxUnsigned w `BV.ult` hiBV || hiBV `BV.ult` loBV
+         then MSP.panic
+                MSP.Symbolic
+                "absValueToBVRange"
+                [ "AbsValue bounds outside declared bitvector width"
+                , "width: " ++ show w
+                , "bounds: [" ++ show lo ++ ", " ++ show hi ++ "]"
+                ]
+         else Just (loBV, hiBV)
 
 -- | Given that the lower @u@ bits of a @w@-bit value lie in @[lo, hi]@,
 -- compute the range of the full @w@-bit value.
@@ -1865,19 +1870,26 @@ absValueToBVRange w av =
 liftInnerRange ::
   NatRepr w ->
   NatRepr u ->
-  Integer -> Integer ->
-  (Integer, Integer)
-liftInnerRange w u lo hi = (lo, hi + 2 ^ natValue w - 2 ^ natValue u)
+  BV.BV u -> BV.BV u ->
+  (BV.BV w, BV.BV w)
+liftInnerRange w u lo hi =
+  let loW = BV.zresize w lo
+      hiW = BV.zresize w hi
+      maxU = BV.zresize w (BV.maxUnsigned u)
+      adjustment = BV.sub w (BV.maxUnsigned w) maxU
+      hiResult = BV.add w hiW adjustment
+  in (loW, hiResult)
 
 -- | Extract an unsigned arithmetic range @(lo, hi)@ from a 'Jmp.SubRange'.
 subRangeToBVRange ::
   NatRepr w ->
   Jmp.SubRange (M.BVType w) ->
-  (Integer, Integer)
+  (BV.BV w, BV.BV w)
 subRangeToBVRange w (Jmp.SubRange rp) =
-  liftInnerRange w (Jmp.rangeWidth rp)
-    (toInteger (Jmp.rangeLowerBound rp))
-    (toInteger (Jmp.rangeUpperBound rp))
+  let u = Jmp.rangeWidth rp
+      lo = BV.mkBV u (toInteger (Jmp.rangeLowerBound rp))
+      hi = BV.mkBV u (toInteger (Jmp.rangeUpperBound rp))
+  in liftInnerRange w u lo hi
 
 -- | Emit 'MacawNarrowBVDomain' expressions for each register in the block's
 -- pre-state that has a non-trivial range, drawing from both the
@@ -1937,10 +1949,10 @@ addAbsValueNarrowingForBlock absSt jumpBounds = do
           (Nothing, Just r) -> Just r
           (Nothing, Nothing) -> Nothing
         guard (lo /= hi)
-        guard (lo /= 0 || hi /= 2 ^ natValue w - 1)
+        guard (lo /= BV.zero w || hi /= BV.maxUnsigned w)
         Just $ do
           orig <- crucibleValue (C.GetStruct startStruct crucIdx repr)
-          narrowed <- evalMacawExt (MacawNarrowBVDomain w' (WUB.range w' lo hi) orig)
+          narrowed <- evalMacawExt (MacawNarrowBVDomain w' (WUB.range w' (BV.asUnsigned lo) (BV.asUnsigned hi)) orig)
           pure (Pair crucIdx narrowed : acc)
 
 addParsedBlock :: forall arch ids s
--- a/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
+++ b/symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
@@ -81,9 +81,11 @@ module Data.Macaw.Symbolic.CrucGen
   , evalAtom
   , crucibleValue
   , freshValueIndex
+  , addAbsValueNarrowingForBlock
+  , absValueToBVRange
   ) where
 
-import           Control.Monad (foldM, forM, unless)
+import           Control.Monad (foldM, forM, guard, unless)
 import           Control.Monad.Except (MonadError(..), ExceptT, runExceptT)
 import qualified Control.Monad.Fail as MF
 import           Control.Monad.State.Strict (MonadState(..), StateT(..), gets, modify')
@@ -92,9 +94,14 @@ import qualified Data.BitVector.Sized as BV
 import qualified Data.Foldable as F
 import           Data.Functor (void, (<&>))
 import qualified Data.Kind as K
+import qualified Data.Macaw.AbsDomain.AbsState as MA
+import qualified Data.Macaw.AbsDomain.JumpBounds as Jmp
+import qualified Data.Macaw.AbsDomain.StackAnalysis as MAS
+import qualified Data.Macaw.AbsDomain.StridedInterval as MASI
 import qualified Data.Macaw.CFG as M
 import qualified Data.Macaw.CFG.Block as M
 import qualified Data.Macaw.Discovery.State as M
+import qualified Data.Macaw.Symbolic.Panic as MSP
 import qualified Data.Macaw.Types as M
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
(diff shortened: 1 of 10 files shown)


langston-barrett:
Oh whoops I forgot I need a What4-side change for this: adding Eq, Ord to the Arith domain

langston-barrett:
I'll wait on https://github.com/GaloisInc/macaw/issues/620 for this.

langston-barrett:
This should also wait on #622, don't want to promote those bugs to symex soundness bugs.

symbolic/src/Data/Macaw/Symbolic/MemOps.hs
```diff
@@ -358,6 +364,63 @@ doPtrToBits sym (Mem.LLVMPointer base off) =
...
+      MSP.panic
+        MSP.Symbolic
+        "narrowBVDomain"
+        [ "Unsoundness! Abstract values are disjoint!"
+        , "width: " ++ show w
+        , "supplied: " ++ show dom
+        , "existing: " ++ show existing
+        , "location: " ++ show (WPL.plSourceLoc loc)
+        , "function: " ++ show (WPL.plFunction loc)
+        ]
```
RyanGlScott, line 399: I suppose this means that any unsoundness in `macaw`'s jump-bound analysis will now be promoted to panics instead of being ignored? Probably the right thing to do, but it does mean that we'll have to be diligent about fixing sources of unsoundness in the future, as they'll block us going forward.

(I assume that you've done some testing of this patch in larger `macaw`-based workflows as a smoke test to catch potential sources of unsoundness that wouldn't have been caught in `macaw`'s test suites?)

symbolic/src/Data/Macaw/Symbolic/MemOps.hs
```diff
@@ -358,6 +364,63 @@ doPtrToBits sym (Mem.LLVMPointer base off) =
...
+      then pure ptr
+      else do
+        (_, off') <- annotateTerm sym (unsafeSetAbstractValue dom' off)
+        pure (Mem.LLVMPointer base off')
+
+-- | Like 'narrowBVDomain', but also adds a proof obligation that the
+-- offset lies within the unsigned bounds of @dom@. Useful in tests to
+-- catch unsoundness in the discovery-time abstract domain by routing
+-- the narrowing through an SMT solver.
+narrowBVDomainChecked ::
```
RyanGlScott: I think I'm missing something. Doesn't `narrowBVDomain` also catch unsoundness, just through different means? Under what circumstances would unsoundness detection à la `narrowBVDomainChecked` be preferable to unsoundness detection à la `narrowBVDomain`?

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -1785,6 +1821,128 @@ addRegUpdateForBlock startAddr = do
...
 
+-- | Project a Macaw 'MA.AbsValue' to an unsigned arithmetic range
+-- @(lo, hi)@, if it carries enough information to do so. Returns
+-- 'Nothing' for 'MA.TopV', 'MA.SomeStackOffset', and other values
+-- that don't correspond to a bounded set of bitvectors.
+absValueToBVRange ::
+  M.MemWidth w =>
+  NatRepr n ->
+  MA.AbsValue w (M.BVType n) ->
+  Maybe (Integer, Integer)
```
RyanGlScott: Does it make sense to represent these bounds as `BV n`s instead of as `Integer`s? The former would have the advantage that you could leverage `bv-sized` to simplify code in certain places (e.g., `maxBnd = 2 ^ natValue w - 1` could become `maxBnd = maxUnsigned w`).

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -244,6 +252,31 @@ data MacawExprExtension (arch :: K.Type)
...
+  -- preserved, and the offset's existing abstract domain is met with the
+  -- supplied one. If the meet is no tighter than the existing domain, the
+  -- offset is left unchanged.
+  --
+  -- This is unsafe in the same sense that
+  -- 'What4.Interface.unsafeSetAbstractValue' is: if the supplied domain does
+  -- not actually contain every possible value of the wrapped offset, downstream
+  -- reasoning becomes unsound. Producers must therefore only emit this when the
+  -- bounds come from a sound static analysis.
+  MacawNarrowBVDomain
```
RyanGlScott, line 271: Perhaps `MacawNarrowBVDomain`'s name should explicitly include the word "unsafe" somewhere to make this potential unsafety more obvious?

symbolic/src/Data/Macaw/Symbolic/MemOps.hs
langston-barrett: > I suppose this means that any unsoundness in macaw's jump-bound analysis will now be promoted to panics instead of being ignored?

Yes, if it's caught by this kind of check. This check won't catch *everything*, as it checks for *disjoint* abstract values. For example, if you had `ite (fresh Bool) 5 7` and one abstract value was the true range `[5, 7]` and the other was unsound and returned `[5]`, this would not catch that. Hence `narrowBVDomainChecked`, which uses the SMT solver to *assert* that the bitvector is in the range, which *would* catch errors like this example.

> Probably the right thing to do, but it does mean that we'll have to be diligent about fixing sources of unsoundness in the future, as they'll block us going forward.

True, we could also consider adding an explicit helper that just ignores these statements in order to make it easier for downstream to mitigate these bugs. Downstream consumers could also provide their own interpretations that just e.g., warn.

> (I assume that you've done some testing of this patch in larger macaw-based workflows as a smoke test to catch potential sources of unsoundness that wouldn't have been caught in macaw's test suites?)

Not yet, I'll test it in GREASE/Screach now!

symbolic/src/Data/Macaw/Symbolic/MemOps.hs
langston-barrett: Yes, see above comment! I will add a similar discussion to the Haddocks.

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
langston-barrett: I'm on the fence about this. On the one hand, it's really not very unsafe at all - we're just saying "if there's a bug in the analysis then you will get bad results". On the other hand, `MacawNarrowBVDomain` *is* exported. So downstream *could* add invalid statements if they want. Then again, they could also fiddle with the offsets in `MacawReadMem`, which would *also* give unsound results, but we don't call that unsafe... What do you think?

RyanGlScott: I don't feel too strongly about it. I wasn't sure exactly how unsafe it was, and I'm happy with your justification for not including "unsafe" in the name.

symbolic/src/Data/Macaw/Symbolic/MemOps.hs
langston-barrett: Passing GREASE/Screach tests here: https://github.com/GaloisInc/grease/pull/696

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -1785,6 +1821,138 @@ addRegUpdateForBlock startAddr = do
...
+    MA.SubValue (n' :: NatRepr n') innerAv
+      | Just (lo, hi) <- absValueToBVRange n' innerAv ->
+          let (loLifted, hiLifted) = liftInnerRange w n' lo hi
+          in Just (loLifted, hiLifted)
+    _ -> Nothing
+  where
+    checkRange lo hi =
+      let loBV = BV.mkBV w lo
+          hiBV = BV.mkBV w hi
+      in if lo < 0 || hi > BV.asUnsigned (BV.maxUnsigned w) || lo > hi
```
RyanGlScott: Rather than calling `asUnsigned` here, couldn't you compare `maxUnsigned` to `hi` using [`ult`](https://hackage-content.haskell.org/package/bv-sized-1.0.7/docs/Data-BitVector-Sized.html#v:ult)? (You could also compare `0`/`loBV` and `loBV`/`hiBV` in a similar way.)

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
```diff
@@ -1785,6 +1821,138 @@ addRegUpdateForBlock startAddr = do
...
+  BV.BV u -> BV.BV u ->
+  (BV.BV w, BV.BV w)
+liftInnerRange w u lo hi =
+  let loInt = BV.asUnsigned lo
+      hiInt = BV.asUnsigned hi
+      -- Compute the adjustment: 2^w - 2^u using BV maxUnsigned values
+      maxW = BV.asUnsigned (BV.maxUnsigned w)
+      maxU = BV.asUnsigned (BV.maxUnsigned u)
+      adjustment = maxW - maxU
+  in (BV.mkBV w loInt, BV.mkBV w (hiInt + adjustment))
```
RyanGlScott: Similarly, I think you could avoid some `asUnsigned` calls here if you use [`add`](https://hackage-content.haskell.org/package/bv-sized-1.0.7/docs/Data-BitVector-Sized.html#v:add)/[`sub`](https://hackage-content.haskell.org/package/bv-sized-1.0.7/docs/Data-BitVector-Sized.html#v:sub) instead of `(+)`/`(-)` over `Integer`s.

RyanGlScott (approved):

symbolic/src/Data/Macaw/Symbolic/CrucGen.hs
langston-barrett: I'll just leave this as-is.

langston-barrett:
I added a `--check-abs-values` flag to GREASE and Screach that uses `narrowBVDomainChecked` and enabled it in their test suites, will wait for their CI to say it works before merging.

PR #622, opened 2026-05-15
base: Fix bugs in `solveLinearDiophantine`, add more tests

langston-barrett:


base/src/Data/Macaw/AbsDomain/StridedInterval/Internal.hs
```diff
@@ -0,0 +1,172 @@
...
+--
+-- > a * x - b * y = c
+--
+-- Assumes @a > 0@, @b > 0@, and @c /= 0@.  Returns 'Nothing' when no such pair
+-- exists; when one does, returns the lex-least solution.
+--
+-- == Proof of correctness
+--
+-- We must show two things: (1) the function returns 'Nothing' exactly when
+-- no box solution exists, and (2) when a box solution exists, the returned
```
RyanGlScott: For those not in the know (like me), what is a "box solution"?

base/src/Data/Macaw/AbsDomain/StridedInterval/Internal.hs
```diff
@@ -0,0 +1,172 @@
...
+-- === Putting it together
+--
+-- Combining Lemmas 1--3, @S@ is nonempty iff @g | c@ /and/ @t_lower <=
+-- t_upper@; and when it is, @S = { (x(t), y(t)) : t_lower <= t <= t_upper
+-- }@.  Since @aq, bq > 0@, both coordinates of (**) are strictly increasing
+-- in @t@, so @t = t_lower@ simultaneously minimizes @x@ and @y@ and yields the
+-- lex-least element of @S@.
+--
+-- The code mirrors this exactly: it returns 'Nothing' when @cr /= 0@ (Lemma 1)
+-- or @t > t_upper@ (Lemma 3), and otherwise @Just (x(t_lower), y(t_lower))@.
```
RyanGlScott: Is Lemma 2 reflected in the code somewhere?

RyanGlScott (commented):
Please separate the changes that fix bugs from the rest of the changes, which move code to an internal module and introduce testing infrastructure.

base/src/Data/Macaw/AbsDomain/StridedInterval/Internal.hs
langston-barrett: Yeah, lemma 2 is what establishes that the whole body of the computation is correct (yields a solution and all solutions are yielded by it). I added `(Lemma 2)` to the end of the last sentence here!

base/src/Data/Macaw/AbsDomain/StridedInterval/Internal.hs
langston-barrett: Ah, it means "solution in the 2D box that bounds `x` and `y`". I made this more explicit!

base/src/Data/Macaw/AbsDomain/StridedInterval/Internal.hs
```diff
@@ -0,0 +1,175 @@
...
+-- === Putting it together
+--
+-- Combining Lemmas 1--3, @S@ is nonempty iff @g | c@ /and/ @t_lower <=
+-- t_upper@; and when it is, @S = { (x(t), y(t)) : t_lower <= t <= t_upper
+-- }@.  Since @aq, bq > 0@, both coordinates of (**) are strictly increasing
+-- in @t@, so @t = t_lower@ simultaneously minimizes @x@ and @y@ and yields the
+-- lex-least element of @S@.
+--
+-- The code mirrors this exactly: it returns 'Nothing' when @cr /= 0@ (Lemma
+-- 1) or @t > t_upper@ (Lemma 3), and otherwise @Just (x(t_lower), y(t_lower))@
```
RyanGlScott: For symmetry with the code, it would be worth phrasing `t > t_upper` as `t_upper < t` here in the comments (or changing the code accordingly).

base/src/Data/Macaw/AbsDomain/StridedInterval/Internal.hs
```diff
@@ -0,0 +1,175 @@
...
+-- (Lemma 2).
+solveLinearDiophantine :: Integer -> Integer -> Integer
+                       -> Integer -> Integer
+                       -> Maybe (Integer, Integer)
+solveLinearDiophantine a b c a_max b_max
+  | cr /= 0      = Nothing  -- Lemma 1
+  | t_upper < t  = Nothing  -- Lemma 3
+  | otherwise    = Just (n * cq + bq * t, m * cq + aq * t)
+  where
```
RyanGlScott, line 138: Maybe:


```suggestion
  | cr /= 0      = Nothing  -- Lemma 1
  | t_upper < t  = Nothing  -- Lemma 3
  | otherwise    = Just (n * cq + bq * t, m * cq + aq * t) -- Lemma 2
```

RyanGlScott (approved):

base/src/Data/Macaw/AbsDomain/StridedInterval/Internal.hs
RyanGlScott: Wow, that diff did _not_ render like I was expecting it to. My suggestion was to put a `-- Lemma 2` comment on the `otherwise` case.

base/src/Data/Macaw/AbsDomain/StridedInterval/Internal.hs
langston-barrett: Good call, thanks!

PR #626, opened 2026-07-01
feat(base): Parse Clang's `-emit-jump-table-sizes-section`

langston-barrett:


x86/tests/x64/test-jumptable-llvm-sizes.c
```diff
@@ -0,0 +1,59 @@
+// End-to-end test for the .llvm_jump_table_sizes section emitted by Clang
+// (>= 20) under -mllvm -emit-jump-table-sizes-section. The Makefile rule
+// passes that flag and links no-pie so the section's R_X86_64_64 entries
+// resolve to literal addresses (the parser reads raw bytes).
```
RyanGlScott, line 4: What happens if you don't pass `-no-pie`? Does `macaw` error out in that case? If so, is this something that could be addressed as future work?

base/src/Data/Macaw/Memory/LLVMJumpTableSizes.hs
```diff
@@ -0,0 +1,150 @@
...
+  in if totalLen `mod` entryBytes /= 0
+       then Left (SectionSizeError totalLen entryBytes)
+       else Right (go bs0)
+
+-- | Read the @.llvm_jump_table_sizes@ section out of an ELF binary, parse
+-- it, and resolve each address against the provided 'Memory'.
+--
+-- Returns a parse error if the section is malformed, otherwise a list of
+-- per-entry warnings (one per unresolvable address) along with a map from
+-- resolved jump-table base addresses to entry counts.
```
RyanGlScott: I assume that "parse error" corresponds to `SectionSizeError`?

base/src/Data/Macaw/Memory/LLVMJumpTableSizes.hs
```diff
@@ -0,0 +1,150 @@
...
+              Elf.ELFDATA2MSB -> BigEndian
+      sectionBytes =
+        [ Elf.elfSectionData sec
+        | (_, sec) <- F.toList (Elf.headerSections ehi)
+        , Elf.elfSectionName sec == sectionName
+        ]
+      wRepr :: AddrWidthRepr w
+      wRepr = case Elf.headerClass (Elf.header ehi) of
+                Elf.ELFCLASS32 -> Addr32
+                Elf.ELFCLASS64 -> Addr64
```
RyanGlScott: You might be able to simplify this with [`elfAddrWidth`](https://github.com/GaloisInc/macaw/blob/67819c525afb122b4e85d98830b2bbf9c79734c7/base/src/Data/Macaw/Memory/ElfLoader.hs#L85-L88).

base/src/Data/Macaw/Memory/LLVMJumpTableSizes.hs
```diff
@@ -0,0 +1,150 @@
...
+  -> Endianness
+  -> BS.ByteString
+  -> Either SectionSizeError [(MemWord w, JumpTableSize w)]
+parseLLVMJumpTableSizes wRepr end bs0 = addrWidthClass wRepr $
+  let ptrBytes = fromIntegral (addrWidthReprByteCount wRepr) :: Int
+      entryBytes = 2 * ptrBytes
+      totalLen = BS.length bs0
+      readWord bs = case wRepr of
+        Addr32 -> fromIntegral (bsWord32 end bs)
+        Addr64 -> bsWord64 end bs
```
RyanGlScott, line 100: Perhaps worth adding a `readWord :: BS.ByteString -> Word64` type signature here to make it clearer from the `fromIntegral` is casting to.

base/src/Data/Macaw/Memory/LLVMJumpTableSizes.hs
```diff
@@ -0,0 +1,150 @@
...
+  deriving Eq
+
+instance PP.Pretty SectionSizeError where
+  pretty = \case
+    SectionSizeError total entry ->
+      PP.pretty (show sectionName) PP.<+> "section size" PP.<+> PP.pretty total
+        PP.<+> "is not a multiple of entry size" PP.<+> PP.pretty entry
+
+instance Show SectionSizeError where
+  show = PP.renderString . PP.layoutPretty PP.defaultLayoutOptions . PP.pretty
```
RyanGlScott: Perhaps this (a bit shorter):


```suggestion
  show = show . PP.pretty
```

And similarly in the `Show` instance for `UnresolvableAddress`.

x86/tests/ElfX64Linux.hs
```diff
@@ -125,7 +126,14 @@ testDiscovery saveMacaw expectedFilename elf = do
...
                               | sym <- syms
                               ]
-  let di = MD.cfgFromAddrs RO.x86_64_linux_info mem addrSymMap entries []
+  (jtWarns, jtSizes) <- case JT.llvmJumpTableSizesFromElf elf mem of
+    Left err -> error $ "Failed to parse .llvm_jump_table_sizes: " ++ show err
+    Right r -> pure r
+  unless (null jtWarns) $
+    error $ "Warnings while parsing .llvm_jump_table_sizes: " ++ show jtWarns
+  let initState = MD.emptyDiscoveryState mem addrSymMap RO.x86_64_linux_info
+                    & MD.llvmJumpTableSizes .~ jtSizes
```
RyanGlScott, line 135: It's somewhat interesting that you have to manually call `llvmJumpTableSizesFromElf` here and then plumb the results through to Macaw's code discovery---will all downstream `macaw` users who want to take advantage of this new feature have to do the same?

x86/tests/x64/test-jumptable-llvm-sizes.c
langston-barrett: You would get `UnresolvableAddress` warnings, as the addresses would be relocations. Noted in a comment, will file an issue for PIE support post-merge!

base/src/Data/Macaw/Memory/LLVMJumpTableSizes.hs
langston-barrett: Yes, thanks!

langston-barrett: Comment updated

base/src/Data/Macaw/Memory/LLVMJumpTableSizes.hs
langston-barrett: Done!

base/src/Data/Macaw/Memory/LLVMJumpTableSizes.hs
langston-barrett: Done!

x86/tests/ElfX64Linux.hs
langston-barrett: Yeah, they will. The issue is that Macaw's discovery is very generic, it doesn't know about ELF. It just takes generic parameters like `Memory` and this jump table size map. Plus, consumers will need to handle errors and warnings in a way that makes sense for their application.

I'm certainly open to suggestions, but this seemed the best default for now!

RyanGlScott (approved):

PR #627, opened 2026-07-01
fix(base): Fix a bug in jump-table analysis involving narrow values

langston-barrett:


RyanGlScott:
I'm not exactly sure what the bug being fixed is here---can you open an issue that describes the misbehavior, and then link this PR back to the issue?

langston-barrett:
The test-case has a description of the bug:
```c
// The bug being guarded against is recording that (32-bit-derived)
// range as a bound on all 64 bits of the register: a value whose low bits
// satisfy the guards but whose high bits are set still reaches the jump table,
// so a 64-bit bound would be unsound.
```
There is also an inline comment:
```haskell
        -- We keep the predicate width @wp@
        -- rather than the add's operation width (@_w@): the bound only covers
        -- the low @wp@ bits of @x@, so promoting it to all of @x@ is unsound
        -- (e.g. a 32-bit compare feeding a 64-bit register).
```
If you revert the fix, you can see the failure mode, which is via `narrowBVDomainChecked`, which emits assertions to the SMT solver that the abstract values that Macaw embeds into Crucible CFGs are correct (that no feasible value can lie outside of them):
```
        tests/pass/switch-jump-table-abs.opt.exe: FAIL (0.49s)
          Testing test_needs_escape (0.46s)
            Failed to prove goal: tests/pass/switch-jump-table-abs.opt.exe:0x401013: error: in test_needs_escape
            narrowBVDomainChecked: offset outside abstract domain bounds
            Details:
              range: [35, 125]
              error, called at src/Data/Macaw/Symbolic/Testing.hs:416:33 in macaw-symbolic-0.0.1-inplace:Data.Macaw.Symbolic.Testing
```
I will put this in an issue and link it from the PR description.

langston-barrett:
See #628, I reference it in the test file and commit message.

RyanGlScott:
Thanks, I was missing the fact that this would trigger a `narrowBVDomainChecked` failure.

RyanGlScott (approved):

